PR #57, opened 2011-06-10
Correctly set the ancestry column when creating a child within an after_* callback

jstirk:
I noticed an issue where creating a child record with the ancestry field set to just the parent id, rather than the entire parent path.

I have added a test reproducing the behaviour, and a fix to the code.

The main issue looked to be that the child_ancestry method was using the ancestry path before the save (using _was) which for new records was obviously blank.

This revision uses the current value of the parent's ancestry path, and adds a child_ancestry_was method to return the old behaviour (with a more descriptive name).

The change to update_descendants_with_new_ancestry was due to it relying on the old behaviour. The revised version explicitly replaces the old parent ancestry snippet (using child_ancestry_was) with the new path.

Please let me know if you need any additional information, or if you would like me to revise the patch.

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -18,8 +18,8 @@ module Ancestry
               descendant.update_attribute(
                 self.base_class.ancestry_column,
                 descendant.read_attribute(descendant.class.ancestry_column).gsub(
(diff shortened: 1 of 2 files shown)


kbrock:
please see #263

PR #99, opened 2012-08-23
I18n support for exception and error strings

anthonybatchelor-examtime:
Added I18n support for all exception and error strings used in ancestry.

This should have very little impact as no `real code` has been changed.


kbrock:
Rebased this PR and moving this discussion over to #253

PR #132, opened 2013-05-02
Added the ability to set the ancestry column delimiter

sholden:
The default delimiter remains `'/'`, but it is now configurable. When using Mysql, setting the delimiter to `','` allows you to use the `FIND_IN_SET` function against the ancestry column, which can be useful in certain situations.

This works correctly against the current version of Rails, but older versions do not allow passing a lambda to `validates_format_of`. Depending on input from other users, we can keep it backward compatible by validating with a custom validator or adding an instance method.

### DIFF
--- a/lib/ancestry.rb
+++ b/lib/ancestry.rb
@@ -4,5 +4,5 @@ require File.join(File.expand_path(File.dirname(__FILE__)), 'ancestry/exceptions
 require File.join(File.expand_path(File.dirname(__FILE__)), 'ancestry/has_ancestry')
 
 module Ancestry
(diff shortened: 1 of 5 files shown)


kbrock:
I'm not sure if this exact change is good, but it is agreed that using an array or set or more tree friendly structure works better.

Looking into this

stefankroes:
If all the tests pass, this change seems fine to me

kbrock:
:+1: this looks good

kbrock:
@stefankroes verify - is this worth dropping older versions of rails? (I'm thinking yes, but want your take)

kbrock:
This is old and will not merge.
I rebased this and produced #262 

Thanks @sholden

PR #183, opened 2014-03-19
allow ids that matches word characters

ghost:
Simple change in the code that allows using string id columns with other than numerical characters. With this change it is pretty easy to create e.g. geography tree that has ISO 3166-1 alpha-3 + regions with custom code. Or language trees, where a root can be e.g. English (en) but its children can have region information like US English (en_US).


kbrock:
Thank you for the PR @ghost

PR #189, opened 2014-05-05
Only propagating updated_at if records have been changed with touch option

brocktimus:
This is a sister PR to the main rails one I've written rails/rails#14979. 

Basically, just adjusting the callback to only run if the record has actually changed. This brings it into line with updated_at on the main record itself which only updates if the record changed.

### DIFF
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -74,9 +74,9 @@ class << ActiveRecord::Base
       }
     end
 
-    after_save :touch_ancestors_callback
     after_touch :touch_ancestors_callback
     after_destroy :touch_ancestors_callback
+    after_save :touch_ancestors_callback, if: :changed?
   end
 end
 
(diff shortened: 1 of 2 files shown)


lib/ancestry/has_ancestry.rb
```diff
@@ -74,9 +74,9 @@ def has_ancestry options = {}
       }
     end
 
-    after_save :touch_ancestors_callback
     after_touch :touch_ancestors_callback
     after_destroy :touch_ancestors_callback
```
mch83: test comment

test/concerns/touching_test.rb
```diff
@@ -44,4 +44,25 @@ def test_touch_option_enabled
       assert_equal way_back, child_1_2.reload.updated_at,        "unrelated record was touched"
     end
   end
-end
\ No newline at end of file
+
```
mch83: test

kbrock:
@stefankroes this looks good to me. Any reason to hold off?

stefankroes:
Nope, should be an improvement

brocktimus:
@kbrock noticed you're now a collaborator so just bumping this a little? (Sorry, not wanting to nag).

PR #191, opened 2014-05-27
Added possibility to set validation regexp pattern in model instead of g...

aderyabin:
We use ancestry in our project for several models. One model uses UUID format as primary key (other uses integer by default). In this case we need to have separate `ANCESTRY_PATTERN` for this model. I suggest to define this constant in model instead of globally declaration.

Let's describe in example:

```ruby
class Category < ActiveRecord::Base
  has_ancestry
end

class GlobalCategory < ActiveRecord::Base
  # lets set pattern for UUID type 
  ANCESTRY_PATTERN = /\A[\w\-]+(\/[\w\-]+)*\z/
  has_ancestry
end
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    ancestry (2.0.0)
+    ancestry (2.1.0)
       activerecord (>= 3.0.0)
 
 GEM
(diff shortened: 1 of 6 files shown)


kbrock:
Thank you for the PR @aderyabin

PR #195, opened 2014-07-25
Implemented very simple caching to avoid unnecessary database queries

brocktimus:
I had some code which was calling `#parent` a lot. Rails `:belongs_to` caches this to avoid hitting the database unnecessarily. After migrating to ancestry I noticed it was much slower because every call to `#parent` resulted in the database being hit.

This is obviously an incredibly naive / trivial cache and doesn't bother to even try to cache the more complex methods.

We could make it an opt in argument passed to `::has_ancestry`, but that seems really conservative.

Any idea how to test this? Can we assert on number of queries run? Otherwise I'll write a test now asserting on the object equality, potentially through `#object_id`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -36,12 +36,6 @@ GEM
       simplecov (>= 0.7)
       term-ansicolor
       thor
-    debugger (1.6.5)
-      columnize (>= 0.3.1)
-      debugger-linecache (~> 1.2.0)
-      debugger-ruby_core_source (~> 1.3.1)
-    debugger-linecache (1.2.0)
-    debugger-ruby_core_source (1.3.1)
     docile (1.1.1)
     i18n (0.6.9)
     json (1.8.1)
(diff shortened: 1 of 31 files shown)


kbrock:
I think this looks neat - but it makes me nervous.
I've seen a number of entity caching layers over the years for rails. They seem like such a good idea until they don't. Rails itself ended up going the caching sql route because they couldn't make it work.

I wonder if there is a way to optionally add this, as I'm leaning away from merging this.
But not against modifying the code to make this extension necessary.

kbrock:
Rails has a standard mechanism for storing relationships in an object. This is cleared out when calling a `reload!`. Seems we should use the rails cache to store the values.

Also, with the introduction of rails 6.0, `children.create` no longer works without deprecations.
So I'm thinking we need to rework the way `children` is setup.

If we can turn it into a relationship, we gain a lot of great stuff. (like `caching` and eager `includes`)

kbrock:
I'm currently trying to convert parent_id and ancestors as standard rails columns. I'm hoping that will lead us to being able to use standard associations (or slightly modified ones)

PR #231, opened 2015-05-13
Allow arrange serializable to accept block

mastfish:
Should make it nice and easy for anyone to use custom serialisation logic.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    ancestry (2.1.0)
+    ancestry (2.0.0)
       activerecord (>= 3.0.0)
 
 GEM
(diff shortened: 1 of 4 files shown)


stefankroes:
Excellent! Looks great. One tiny pet peeve: Could you rename childs to children in tests and especially docs? If you are not willing to I'll merge as is but it should be a small effort.

mastfish:
Sure thing - done.

stefankroes:
Merged. Thank you for contributing!

defbyte:
hello there!

I noticed this commit hasn't made its way in to the published Gem. It is present on `2-1-stable` but not the tagged `2.1.0` release. Just wondering when this change will make its way in to the published Gem.

thanks!

PR #238, opened 2015-08-11
Scope for leaves

antstorm:
Hi there,

Here's a pure-SQL support for getting all the leaves of the tree. Basically making sure that for every record there's other record that references it's `ancestry`. I had to use `OR` in the `JOIN` because `ancestry` might be nil.

### DIFF
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -38,6 +38,7 @@ class << ActiveRecord::Base
 
     # Named scopes
     scope :roots, lambda { where(ancestry_column => nil) }
+    scope :leaves, lambda { joins("LEFT JOIN #{table_name} AS c ON c.#{ancestry_column} = CAST(#{table_name}.id AS text) OR c.#{ancestry_column} = #{table_name}.#{ancestry_column} || '/' || #{table_name}.id").group("#{table_name}.id").having('COUNT(c.id) = 0') }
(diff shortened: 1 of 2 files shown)


antstorm:
Dammit MySQL…

kbrock:
yea, mysql likes '+' vs '||'.

you may want to try 'concat' - it is not ANSI but looks to have better support. at least for more recent postgres/sqlite versions.

GL

antstorm:
@kbrock it doesn't seem like SQLite has `concat` support, whichever you use there's always an outlier. I think I'll just have to use DB detection… ugly, but so is the whole situation with `concat`.

antstorm:
Took a different approach and reopened this PR — #247

BjarneD:
@ImDineshSaini 
I dont know your exact error, but i also had an error. If you use mysql as I do on this project, you can use this solution.

There appeared to be two issues on my end. Concatenation and cast method.
Instead of || or + I used mysql concat function
Cast method doesn't support type `TEXT` so I use char(50)

Here is my mysql solution
`scope :leaves, lambda { joins("LEFT JOIN #{table_name} AS c ON c.#{ancestry_column} = CAST(#{table_name}.id AS char(50)) OR c.#{ancestry_column} = concat(#{table_name}.#{ancestry_column}, '/', #{table_name}.id)").group("#{table_name}.id").having('COUNT(c.id) = 0') }`

PR #239, opened 2015-08-11
Ignore locked gemfiles

antstorm:
`Gemfile.lock` and other locked gemfiles shouldn't be in the repo — http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/


kbrock:
I think we need to lock some of the gem versions in rails 3.*.
Running many of these locally work for me but fail in travis. very odd

kbrock:
Thanks.

rebased / Merged the git ignore.

antstorm:
Cool, thanks!

PR #245, opened 2015-12-03
Range children

kbrock:
The current code is generating case insensitive searches on postgres rails 4.2

It now uses ranges, which will work well for indexes that support ranges

```sql
-- postgres index
create index index_users_on_ancestry on users (ancestry varchar_pattern_ops);
```

Sorry to push code not tested on mysql, having trouble getting the test suite working.


kbrock:
oh well. very confused. These all pass for me locally.

name|version||
---|---|---
osx | 10.10.5
ruby |1.9.3-p547|2.1.5
postgress|9.2.8
mysql|5.6.26
sqlite|3.9.2

These all pass for me:

```
BUNDLE_GEMFILE=gemfiles/pg_ar_32.gemfile bundle exec rake
BUNDLE_GEMFILE=gemfiles/pg_ar_40.gemfile bundle exec rake
BUNDLE_GEMFILE=gemfiles/mysql_ar_40.gemfile bundle exec rake
```

Wasn't able to run these under ruby 2.2 - it just ignored the tests.

I'll look into this further

antstorm:
That's really odd, because it looks like a legit test failure in Travis. It might be worth putting some kind of debug output in there just to make sure that queries on your local machine match the ones generated in Travis.

kbrock:
Was able to get this running, but between doesn't seem to buy much. only thing to buy is that case sensitive seems a little bit faster.

kbrock:
fun followup:

I'm noticing that mysql and postgres translates `LIKE` statements to `BETWEEN` (well, a `~~>=~~` and a `~~<~~`).

So we definitely were onto something and it is nice that the database is doing it for us.

PR #246, opened 2015-12-05
Scope for leaves

antstorm:


kbrock:
I'd too would like to add postgres specific version of some other methods. e.g.: sorting by nulls vs using a case statement.

At a hight level, is the code here the best way to introduce database specific sql?

antstorm:
@kbrock I'd say if more methods start to appear then it's better to extract them into an `SQLHelper` module or something. Would be great for adapters to support that, but I think it might be out of their scope.

kbrock:
Hi @antstorm It is a long shot, but seeing if we can put this abstraction into arel [here](https://github.com/rails/arel/pull/410)

antstorm:
@kbrock good shout, let's see what the arel guys think. Judging by the previous PR attempt they're not very positive about supporting infix operators.

antstorm:
@kbrock nice one, I never thought it would make it's way into `arel`! Although I don't think I can use that here, since ancestry depends on `activerecord >= 3.0.0`.

ImDineshSaini:
Alternate is add a file in config/initializer 

```
module Ancestry
  module ClassMethods
    # Scope that returns all the leaves
    def leaves
      id_column = "#{table_name}.id"
      id_column_as_text = sql_cast_as_text(id_column)
      parent_ancestry = sql_concat("#{table_name}.#{ancestry_column}", "'/'", id_column_as_text)

      joins("LEFT JOIN #{table_name} AS c ON c.#{ancestry_column} = #{id_column_as_text} OR c.#{ancestry_column} = #{parent_ancestry}").
        group(id_column).
        having('COUNT(c.id) = 0')
    end

    private

    def sql_concat *parts
      if ActiveRecord::Base.connection.adapter_name.downcase == 'sqlite'
        parts.join(' || ')
      else
        "CONCAT(#{parts.join(', ')})"
      end
    end

    def sql_cast_as_text column
      text_type = if ActiveRecord::Base.connection.adapter_name.downcase == 'mysql2'
        'CHAR'
      else
        'TEXT'
      end

      "CAST(#{column} AS #{text_type})  collate utf8_unicode_ci"
    end
  end
end
```

kbrock:
@antstorm Arel supports `concat` (see reference above to rails PR)
I'd like to release ancestry v3 with rails 4.x support and above.
And then start to use stuff like this.

Ancestry 2.1.x can support older rails, just not get these newer features.
Are you still using ancestry? What version of rails do you need to support?

antstorm:
@kbrock I'm not using `ancestry` in my current projects, but I'm happy to shape this PR into something mergeable, dropping `rails < 4` support.

fursich:
Hi @antstorm , would you be comfortable if I rebase this PR and give it a new life, since I found it much better than mine (#388)  in terms of SQL efficiency and simplicity.
As a daily user of ancestry I appreciate the efforts made here, and leaving this feature un-merged is just a shame..

(@kbrock I'm assuming currently nobody is working on it, let me know if I'm in a totally wrong place.. you might already have clear ideas about what to be done.)

Hope that makes sense 😸

antstorm:
@fursich sure, I've rebased it. Do you think there's anything missing from it?

.travis.yml
```diff
@@ -36,5 +36,5 @@ services:
   - postgresql
 
 before_script:
-  - mysql -e 'create database ancestry_test;' || true
-  - psql -c 'create database ancestry_test;' -U postgres || true
+  - mysql -e 'CREATE DATABASE ancestry_test CHARACTER SET utf8 COLLATE utf8_general_ci;' || true
```
kbrock: Hi @antstorm 

I think this line is significant. Could you please explain more about this?

We may want to add this to the general documentation to help people.
Also if this is not changed as the database collation, then I wonder what users can do to get this behavior.

I'm also thinking postgres on mac has a similar issue with this.

lib/ancestry/class_methods.rb
```diff
@@ -17,6 +17,17 @@ def scope_depth depth_options, depth
       end
     end
 
+    # Scope that returns all the leaves
+    def leaves
+      id_column = "#{table_name}.id"
```
kbrock:
```suggestion
      id_column = "#{table_name}.#{self.class.primary_key}"
```

lib/ancestry/class_methods.rb
```diff
@@ -17,6 +17,17 @@ def scope_depth depth_options, depth
...
+      id_column_as_text = sql_cast_as_text(id_column)
+      parent_ancestry = sql_concat("#{table_name}.#{ancestry_column}", "'/'", id_column_as_text)
+
+      joins("LEFT JOIN #{table_name} AS c ON c.#{ancestry_column} = #{id_column_as_text} OR c.#{ancestry_column} = #{parent_ancestry}")
+        .group(id_column)
+        .having('COUNT(c.id) = 0')
```
kbrock:
```suggestion
        .having("COUNT(c.#{self.class.primary_key}) = 0")
```

lib/ancestry/class_methods.rb
```diff
@@ -17,6 +17,17 @@ def scope_depth depth_options, depth
       end
     end
 
+    # Scope that returns all the leaves
+    def leaves
+      id_column = "#{table_name}.id"
+      id_column_as_text = sql_cast_as_text(id_column)
+      parent_ancestry = sql_concat("#{table_name}.#{ancestry_column}", "'/'", id_column_as_text)
+
+      joins("LEFT JOIN #{table_name} AS c ON c.#{ancestry_column} = #{id_column_as_text} OR c.#{ancestry_column} = #{parent_ancestry}")
```
kbrock: this looks like "children" to me.
Think we may want "subtree" instead?

test/concerns/scopes_test.rb
```diff
@@ -6,6 +6,9 @@ def test_scopes
       # Roots assertion
       assert_equal roots.map(&:first), model.roots.to_a
 
+      # Leaves assertion
+      assert_equal model.all.select(&:is_childless?), model.leaves.order(:id).to_a
```
kbrock: well, this does look like it is bringing back nodes from 2 levels deep.

ASIDE (nothing to do with you or this PR):
The way the tests are organized, it sure makes it hard to know where different elements are tested.

lib/ancestry/class_methods.rb
kbrock: but it seems to be working.

Could you just give your opinion on what you think about this line and multiple levels of nested members?

Thnx

iainbeeston:
Is there any chance of resurrecting this? What else needs doing for it to be mergeable?

.travis.yml
kbrock: or maybe this line should be collation on just the column if they don't do this to the whole database?

kbrock:
Do you want to put together another PR?

We now have a postgres only mixin, so we could do the same style for mysql and postgres to get past the database specific concat syntax. Which I think was the only issue.

I did find an old gem called `ancestry_joins` which seems to do similar stuff like this as well. Maybe we could compare the 2 approaches and find one that works best for us

.travis.yml
kbrock: I added collation to the column and not to the whole database.
Thanks for putting this on my radar.

kbrock:
@iainbeeston sorry for the delay here.

Yes, this is still on my radar.
I'm working at finding a way to drop `to_node` so we can use scopes in our queries and make this work well.

kbrock:
thank you. #725

PR #251, opened 2015-12-24
Minitest

kbrock:
Use mini-test


kbrock:
looks like this will be tricky since the rails testing api changed from 3.0 to 4.0

kbrock:
fixed in #289

PR #253, opened 2016-01-01
I18n support for exception and error strings

kbrock:
fixes #99 from @anthonybatchelor-examtime

Does this still look good?

### DIFF
--- a/lib/ancestry.rb
+++ b/lib/ancestry.rb
@@ -6,6 +6,9 @@ require_relative 'ancestry/has_ancestry'
 require_relative 'ancestry/materialized_path'
 require_relative 'ancestry/materialized_path_pg'
 
+I18n.load_path += Dir[File.join(File.expand_path(File.dirname(__FILE__)),
+                                 'ancestry', 'locales', '*.{rb,yml}').to_s]
+
 module Ancestry
   @@default_update_strategy = :ruby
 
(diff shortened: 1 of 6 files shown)


stefankroes:
IMHO this isn't useful for the majority of the errors that where localized as they are means to be developer facing, not user facing. Where errors are user facing, I agree that this is useful

kbrock:
Revisiting this.

Most of these errors look like they are developer centric.
Not sure which of these (if any) need to be displayed to the end user.

It did get a number of +1 requests though.

I'll rebase and see how this fairs

brunowego:
@stefankroes please merge this! Thanks @kbrock 👍

kbrock:
kicking

kbrock:
Would having an error code in the exceptions be better than displaying the error to the user.

Seems most of these errors are due to developer error rather than user error.
So using localization seems odd

cesarjr:
Is there any change this pull request to be merged?

kbrock:
@cesarjr I just rebased this PR and it is ready for merge

All of these messages seem to be developer focused.
And typically the purpose of I18N is for updating messages to the end user in the web interface.

A lot of people are asking for this, so I understand that it is desired.
But what do you hope to get with localized error messages since most will not be displayed to the end user?
I'd like to know so I can make sure it meets people's needs.

cesarjr:
Hi!

I'm gonna give you my example.

My model is set to `has_ancestry orphan_strategy: :restrict`.

When the user try to destroy a node with children, ancestry is gonna raise the `Ancestry::AncestryException`.

At the controller I have to rescue this error, check if the message has `/it has descendants/` and send a new message in my language (Portuguese) to the user.

if the message could be translated I just could get the error message and deliver to the user directly.

This is why I think this PR useful in my case.

What do you think about that?

d-m-u (approved):
i think this looks great, and i think we should have it.

PR #254, opened 2016-01-13
altered ANCESTRY_PATH to accept a / separated list of UUID strings

stevegoobermanhill:
Hi Stefan,
I've found that ancestry wouldn't work with UUID primary keys as defined using the ActiveUUID gem (which is quite nice). The issue appears to be that the ANCESTRY_PATTERN regexp is looking for a / separated list of integer primary keys - when the  UUID primary keys are UUID strings.

I've patched the ANCESTRY_PATTERN regexp so it will accept UUID primary keys.

Kind regards

Steve

### DIFF
--- a/lib/ancestry.rb
+++ b/lib/ancestry.rb
@@ -4,5 +4,11 @@ require_relative 'ancestry/exceptions'
 require_relative 'ancestry/has_ancestry'
 
 module Ancestry
-  ANCESTRY_PATTERN = /\A[0-9]+(\/[0-9]+)*\Z/
+  ANCESTRY_PATTERN = %r{
+                        \A
+                        ([0-9]+(\/[0-9]+)*) #integers separated by / 
+                        |                   #or a list of UUID strings separated by /
+                        ((?<uuid>\p{XDigit}{8}-\p{XDigit}{4}-\p{XDigit}{4}-\p{XDigit}{4}-\p{XDigit}{12})(\/\g<uuid>)*)
+                        \Z     
+                     }x
 end


kbrock:
Hello @stevegoobermanhill 

Supporting string primary keys would be great.
The key is to have a key delimiter in the ancestry field that is not allowed in the primary key. We currently use "/"

Do you have an idea for extracting `ANCESTRY_PATTERN` into a configuration variable?
We could say that that this needs to be configured before any models are loaded (so no ancestry columns would have been configured yet)

Thanks

stevegoobermanhill:
Hi Keenan
Supporting generic string primary keys is a little more difficult than supporting UUIDs – which have a known and fixed format.

The way I have done this sort of thing before is to set a class variable (and it has to be a separate class variable in each receiving class of the Ancetry module) to hold the ancestry path separator – which is the current separator (‘/’) by default.

In that way you could generically use strings as primary keys, with the caveat being that you need to have a well-defined separator that isn’t going to conflict with your primary key string format

Another possibility would be to hold the ancestry path as a serialized Array – say using JSON. You then dump and load it as required.

Kind regards

Steve

kbrock:
I've wanted to use postgres arrays for the keys instead of our manipulation. would work very well. But others are pointing me at ltree implementations and I haven't gotten my mind around those

stevegoobermanhill:
My £0.02 worth would be that I would try and keep Ancestry as a purely ruby Gem – if you put Postgres specific dependencies into it, then you are locked into a single DB solution.

Personally I currently use MySQL (MariaDB) and SQLite a lot (admittedly mainly MariaDB for rails projects, once they are established). However, our main software development effort is currently being ported from a monolithic Rails / MariaDB system into a microservices architecture (still using Rails) sitting over a Cassandra NoSQL DB). So I think that keeping Ancestry agnostic with respect to the underlying DB is preferable.

If you use a JSON array to hold the Ancestry keys, then you should be on safe ground because pretty much every DB (possibly not SQLite) has a native JSON type, and even if they don’t, parsing an array to/from JSON has very low overhead.

Steve

kbrock:
Thank you for the PR @stevegoobermanhill 

Appreciate the feedback. It has been a push to get this fixed.

PR #256, opened 2016-01-17
Update .travis.yml

yui-knk:
This PR includes these update

* Update Ruby 2.2.2 -> 2.2.4
* Add Ruby 2.3
* Fix exclude test matrix (we should exclude 1.9.3 not 1.9.2)
* Add exclude test matrix (Ruby 2.2 or 2.3 x Rails3)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,8 @@ sudo: false
 
 rvm:
   - 1.9.3
-  - 2.2.2
+  - 2.2.4
+  - 2.3.0
 
 gemfile:
   - gemfiles/sqlite3_ar_30.gemfile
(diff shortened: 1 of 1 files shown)


kbrock:
@yui-knk Thanks. I like the idea of upgrading ruby to 2.2.4, and excluding older rails on newer rubys.

Is there a particular reason why you want to add tests for ruby 2.0, and 2.1?

yui-knk:
I think if we test against ruby 1.9 and 2.2, it is better to test also 2.0 and 2.1. I feel including all versions of ruby for test targets is natural.

kbrock:
@yui-knk There are just so many combinations, it is distracting and takes so long.
I can see people running ruby 1.9 (kinda) and 2.2, but have trouble believing they are actually running 2.1. And if they are, are they really upgrading this gem?

yui-knk:
I feel taking long time to test is better than no test (for some ruby versions). But I also am agree with long time tests irritates us, so I can remove ruby 2.0 and 2.1. Are you agree with adding 2.3 ?

kbrock:
@yui-knk yes please. that sounds great.

So many ruby versions x so many active record versions.
I feel it is more likely that people use ancestry with the various active record versions.

@stefankroes any ideas on reducing the number of variations here?

yui-knk:
Remove 2.0 and 2.1 :)

PR #259, opened 2016-02-06
use null sorting

kbrock:
Use a function for sorting (which is indexable) over the case statement.


kbrock:
I wasn't expecting this to pass. I'll have to double check the tests

kbrock:
this is passing because postgres, mysql, and sqlite all support coalesce. Looks like it is pretty standard across most databases. Any concerns about compatibility here?

kbrock:
added in check to make sure it running correct code against correct versions.

Only running the check code 1 time at class definition time

PR #261, opened 2016-02-06
Improve performance of arrange_nodes.

Fryguy:
The old code is essentially an insertion algorithm that walks the tree
from the root for each node and inserts itself in the right spot.
Inserting a node is roughly O(depth); best case O(1), but worst case
O(n).  Thus, the tree building is roughly O(n * depth); best case O(n),
but worst case O(n^2).

The new code maintains an index of a node's children by node id.  Each
nodes is placed into the children of the parent id. Thus, inserting a
node is an O(1) lookup, making the tree building O(n).

The operation only has to look at the parent, and does not have to go
through all of the ancestry ids, because eventually the parent node
will go through the same insertion process itself.

Note that the old algorithm used a pre-sorting of nodes in order to
more efficiently arrange them.  The new algorithm doesn't care about
order, and thus that sorting can probably be removed in future PRs.

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -42,18 +42,23 @@ module Ancestry
     # Arrange array of nodes into a nested hash of the form
     # {node => children}, where children = {} if the node has no children
     def arrange_nodes(nodes)
-      # Get all nodes ordered by ancestry and start sorting them into an empty hash
-      nodes.inject(ActiveSupport::OrderedHash.new) do |arranged_nodes, node|
-        # Find the insertion point for that node by going through its ancestors
-        node.ancestor_ids.inject(arranged_nodes) do |insertion_point, ancestor_id|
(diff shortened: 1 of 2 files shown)


Fryguy:
This is marked as WIP as I'd like to test against some real world cases and possibly create some more tests.

kbrock:
@Fryguy looks good. let me know when you are ready for me to pull the trigger

Fryguy:
@kbrock This should be good to go now.

Fryguy:
Actually, I have a way to make the tests readable, so hold off until I push new specs.

Fryguy:
@kbrock Ok, I updated the specs to DRY them up a bit...

kbrock:
@Fryguy thanks

> Ok, I updated the specs to DRY them up a bit...
> @Fryguy 

a bit? looks like you got them down a lot. much nicer to read. thanks

Issue #221, opened 2015-01-21
Primary id column restriction? Please confirm


solars:
Hi,

I've got an application where the primary ID is not .id but a different field as the primary ID is something like FU_BAR_32.

This is perfectly fine and valid in rails and you can define it with self.primary_key =

However, it seems as if ancestry is restricted to the .id field, can you please confirm this?

stefankroes:
I think I merged a patch at some point that updated ancestry to respect the configured primary key field.

solars:
Hm, I have version 2.1.0 - is it too old then?

solars:
There seem to be a lot of conflicting statements about custom primary keys

the primary_key_format option does not work apparently
this also dose not work

I have a model ArProperty with self.primary_key = "property_id"
The primary keys are strings like AB_DE9 etc - how can I specify this for ancestry? Cannot find out..

Issue #241, opened 2015-09-03
Gem does not support primary_key_format option


shaggyone:
Gem declares to support :primary_key_format option but actually uses Ancestry::ANCESTRY_PATTERN constant

nepalez:
The same problem [arising from here](https://github.com/stefankroes/ancestry/blob/2fa4381e0e21289d6800772d7f67d16efa65c4f8/lib/ancestry/has_ancestry.rb#L34)

Spone:
Here is the recommended monkey-patch, until this option is supported: https://github.com/stefankroes/ancestry/wiki/Usage-with-UUID's

kbrock:
is this working for you all without the patch?
Can I close this?

halo:
> is this working for you all without the patch?

Yes, using the `primary_key_format` option works for UUIDs (tried on `4.1.0`).

> Can I close this?

I believe this issue can be closed. Thank you very much!

PR #262, opened 2016-02-06
set the ancestry column delimiter and id pattern

kbrock:
This is based upon #132 by @sholden 

It had some conflicts, so I brought up to date and took a few liberties. I can change attribution if you don't feel it reflects your intent / code.
1. set the `:ancestry_delimiter`
2. set the `:primary_key_format`

This should fix https://github.com/stefankroes/ancestry/issues/221 and https://github.com/stefankroes/ancestry/issues/241

Similar to #265

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -29,8 +29,7 @@ module Ancestry
 
     # Delimiter writer
     def ancestry_delimiter= ancestry_delimiter
-      # Check value of the delimiter. Restrict if it's numeric.
-      if ancestry_delimiter =~ /^\D+$/
+      if ancestry_delimiter !~ /\A#{primary_key_format}\Z/
         class_variable_set :@@ancestry_delimiter, ancestry_delimiter
       else
         raise Ancestry::AncestryException.new("Invalid delimiter, only non-numeric characters are allowed.")
@@ -38,7 +37,7 @@ module Ancestry
     end
 
     def ancestry_pattern
-      /\A[0-9]+(#{Regexp.escape(ancestry_delimiter)}[0-9]+)*\Z/
+      /\A#{primary_key_format}(?:#{Regexp.escape(ancestry_delimiter)}#{primary_key_format})*\Z/
     end
 
     # Arrangement
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -27,6 +27,20 @@ module Ancestry
       end
     end
 
+    # Delimiter writer
+    def ancestry_delimiter= ancestry_delimiter
+      # Check value of the delimiter. Restrict if it's numeric.
+      if ancestry_delimiter =~ /^\D+$/
+        class_variable_set :@@ancestry_delimiter, ancestry_delimiter
+      else
+        raise Ancestry::AncestryException.new("Invalid delimiter, only non-numeric characters are allowed.")
+      end
+    end
+
+    def ancestry_pattern
+      /\A[0-9]+(#{Regexp.escape(ancestry_delimiter)}[0-9]+)*\Z/
+    end
+
     # Arrangement
     def arrange options = {}
       scope =
@@ -168,7 +182,7 @@ module Ancestry
             # ... rebuild ancestry from parents array
             ancestry, parent = nil, parents[node.id]
             until parent.nil?
-              ancestry, parent = if ancestry.nil? then parent else "#{parent}/#{ancestry}" end, parents[parent]
+              ancestry, parent = if ancestry.nil? then parent else "#{parent}#{ancestry_delimiter}#{ancestry}" end, parents[parent]
             end
             node.without_ancestry_callbacks do
               node.update_attribute node.ancestry_column, ancestry
@@ -185,7 +199,7 @@ module Ancestry
           node.without_ancestry_callbacks do
             node.update_attribute ancestry_column, ancestry
           end
-          build_ancestry_from_parent_ids! node.id, if ancestry.nil? then "#{node.id}" else "#{ancestry}/#{node.id}" end
+          build_ancestry_from_parent_ids! node.id, if ancestry.nil? then "#{node.id}" else "#{ancestry}#{ancestry_delimiter}#{node.id}" end
         end
       end
     end
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -2,16 +2,13 @@ class << ActiveRecord::Base
   def has_ancestry options = {}
     # Check options
     raise Ancestry::AncestryException.new("Options for has_ancestry must be in a hash.") unless options.is_a? Hash
-    options.each do |key, value|
-      unless [:ancestry_column, :orphan_strategy, :cache_depth, :depth_cache_column, :ancestry_delimiter, :touch].include? key
-        raise Ancestry::AncestryException.new("Unknown option for has_ancestry: #{key.inspect} => #{value.inspect}.")
-      end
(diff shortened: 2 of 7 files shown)


lib/ancestry/instance_methods.rb
```diff
@@ -321,7 +321,7 @@ def unscoped_descendants
     # Validates the ancestry, but can also be applied if validation is bypassed to determine if children should be affected
     def sane_ancestry?
       ancestry_value = read_attribute(self.ancestry_base_class.ancestry_column)
-      ancestry_value.nil? || (ancestry_value.to_s =~ Ancestry::ANCESTRY_PATTERN && !ancestor_ids.include?(self.id))
+      ancestry_value.nil? || (ancestry_value.to_s =~ self.ancestry_base_class.ancestry_pattern && !ancestor_ids.include?(self.id))
```
kbrock: > Do you need to cache `ancestry_pattern` since this will be called often?
> @Fryguy 

`sane_ancestry?` should only be called when a node is moved. So this shouldn't be too bad.

Pretty sure the validation only reads `ancestry_pattern` at validator creation.

to be honest, wonder if sane_ancestry can be dumped and just call `valid?` instead

lib/ancestry/instance_methods.rb
```diff
@@ -100,7 +100,7 @@ def ancestry_changed?
     end
 
     def parse_ancestry_column obj
-      obj.to_s.split('/').map { |id| cast_primary_key(id) }
+      obj.to_s.split(self.ancestry_base_class.ancestry_delimiter).map { |id| cast_primary_key(id) }
```
Fryguy: Should this call get_ancestry_delimiter?

lib/ancestry/instance_methods.rb
```diff
@@ -90,7 +90,7 @@ def child_ancestry
       # New records cannot have children
       raise Ancestry::AncestryException.new('No child ancestry for new record. Save record before performing tree operations.') if new_record?
 
-      if self.send("#{self.ancestry_base_class.ancestry_column}_was").blank? then id.to_s else "#{self.send "#{self.ancestry_base_class.ancestry_column}_was"}/#{id}" end
+      if self.send("#{self.ancestry_base_class.ancestry_column}_was").blank? then id.to_s else "#{self.send "#{self.ancestry_base_class.ancestry_column}_was"}#{self.ancestry_base_class.ancestry_delimiter}#{id}" end
```
Fryguy: Should this call get_ancestry_delimiter?

lib/ancestry/instance_methods.rb
```diff
@@ -38,7 +38,7 @@ def apply_orphan_strategy
           if self.ancestry_base_class.orphan_strategy == :rootify
             unscoped_descendants.each do |descendant|
               descendant.without_ancestry_callbacks do
-                descendant.update_attribute descendant.class.ancestry_column, (if descendant.ancestry == child_ancestry then nil else descendant.ancestry.gsub(/^#{child_ancestry}\//, '') end)
+                descendant.update_attribute descendant.class.ancestry_column, (if descendant.ancestry == child_ancestry then nil else descendant.ancestry.gsub(/^#{child_ancestry}#{Regexp.escape(self.ancestry_base_class.ancestry_delimiter)}/, '') end)
```
Fryguy: Should this call get_ancestry_delimiter?

lib/ancestry/class_methods.rb
```diff
@@ -27,6 +27,20 @@ def orphan_strategy= orphan_strategy
       end
     end
 
+    # Delimiter writer
+    def ancestry_delimiter= ancestry_delimiter
+      # Check value of the delimiter. Restrict if it's numeric.
```
Fryguy: This comment no longer makes sense if the pattern can change

lib/ancestry/class_methods.rb
```diff
@@ -27,6 +27,20 @@ def orphan_strategy= orphan_strategy
...
+      if ancestry_delimiter !~ /^#{ancestry_id_pattern}$/
```
Fryguy: This should probably be `/\A#{ancestry_id_pattern}\z/` so as not to match around newlines.

lib/ancestry/has_ancestry.rb
```diff
@@ -31,7 +36,7 @@ def has_ancestry options = {}
     self.touch_ancestors = options[:touch] || false
 
     # Validate format of ancestry column value
-    validates_format_of ancestry_column, :with => Ancestry::ANCESTRY_PATTERN, :allow_nil => true
+    validates_format_of ancestry_column, :with => ancestry_pattern, :allow_nil => true
```
Fryguy: This is where ancestry_pattern is used that I thought would be expensive if calculated over and over.

kbrock: yea, if this were `:ancestry_pattern` then it would call each time.

since this is `ancestry_pattern` pretty sure it evaluates only once before even going into `validates_format_of`

Fryguy: ohhhh interesting...very subtle

Fryguy: So if the pattern is changed after the model is loaded, it will pick it up?

kbrock: @Fryguy no. it will not pick it up.

Was thinking about making the getter/setter for the id and delimiter private for this reason. thoughts?

lib/ancestry/instance_methods.rb
kbrock: @stefankroes I tried to follow the convention in place for whether calling `ancestry_base_class.X` vs `get_X`, but I don't understand the reason for the 3 different ways to access the same values.

Also, the class level variables goes against allowing multiple trees per node, which you had expressed interest. (note: I did not say multiple parents per node.)

I'm thinking about going through and consolidating the number of different ways these are accessed. any comments or advice?

lib/ancestry/has_ancestry.rb
Fryguy: Then how would people set it?

kbrock: There are a few examples in the [tests](https://github.com/stefankroes/ancestry/pull/262/files#diff-55b6e672c474d6ae1a8037ac00471dbfR37).
The thought is that the delimiter would not change after the column is configured.

lib/ancestry/instance_methods.rb
kbrock: line:38 had `ancestry_base_class.orphan_strategy`, so I just used that as a basis.

yea, consistency could be a little better

mmahalwy:
still open?

lib/ancestry/has_ancestry.rb
kbrock: @fryguy do you think we have a lot to flush this out?

voltechs:
Merge when? Massively limiting in functionality currently without this.

kbrock:
This works, but changes too much code all over.
The problem is that too much code is concerned with the delimiter

Want to move the manipulation of the ids into a common place so fewer areas of code are concerned with the delimiter

kbrock:
UPDATE:

The current code requires just too much munging.
I changed the code to move most of the parsing and string concatenation into 1 file (4-5 places)
This should allow this feature and others to work.

Have plans for supporting `'.'` as a delimiter so LTREE will work.

kbrock:
revisit after https://github.com/stefankroes/ancestry/pull/383 merges

Either the delimiter will be in fewer places or the pattern may become moot

kbrock:
this just got too complicated

will remove that from the documentation.

we need to support uuid / string / integer.

far simpler to just let the user setup a validator here, or pass the validator in.
the sanity check can just use `valid?`

Issue #33, opened 2010-11-28
Accessing node's path in before_save or after_save


ghost:
Hello,

I am using ancestry with friendly_id and attempting to write a full path of the node, with it's name seperated by slashes. I am doing this so I can do a single lookup when using Rails route globbing instead of multiple queries.

So far, path always seems to be empty. I've tried to do an AR find from irb after setting a breakpoint in the before_save method and it seems that all queries are scoped to the current records ancestry field contents.

I am not highly skilled in rails, so I might be missing something, or there might be a way to call AR without the scoping, but I am at a loss.

Hopefully you'd be able to suggest a solution. I love ancestry and it seems to fit will for this project and I'd love to move forward!

Sincere,
Nick

tilo:
@stefankroes I'm running into the same issue with ancestry 2.1.0 , Rails 4.2.4 with a Postgres DB.

Can you please re-open this issue, or comment what might be causing this?

```
class Node < ActiveRecord::Base
  has_ancestry(:orphan_strategy => :adopt)
  # has a name and a slug string attribute, the slug is supposed the be the path's names concatenated

  after_save :generate_slug

  # 
  def generate_slug
    puts "OBJECT: #{self.name} ; PATH IS: #{self.path.pluck(:name).inspect}"
    self.update_column(:slug , self.path.each_with_object([]) {|x,a| a << x.name}.join('/') )
  end
end
```

Objects are created in this fashion:

```
root = Node.create(:name => 'root')
child1 = root.children.create(:name => 'child1')
OBJECT: child1 ; PATH IS: ["child1"]
child1.slug
  => 'child1'    # oops!
child1.save
child1.slug
  => 'root/child1'
```

kbrock:
I really didn't like the way the ids were being updated before. Hoping the newer code will work much better for everyone

PR #263, opened 2016-02-08
Fix creating children in after_* callbacks

kbrock:
**NOTE:** REVISED for 2.1.x version of ancestry #57

Previously creating children in an after_\* callback would cause the
child's ancestry column to contain just the parent id and not the entire
ancestry path.

A workaround was to call self.reload in the after_\* callback, but this
commit makes the process work automatically.

The main issue looked to be that the child_ancestry method was using the
ancestry path before the save (using _was) which for new records was
obviously blank.

This revision uses the current value of the parent's ancestry, and adds
a child_ancestry_was method to return the old behaviour (with a more
descriptive name).

The change to update_descendants_with_new_ancestry was due to it relying
on the old behaviour. The revised version explicitly replaces the old
parent ancestry snippet (using child_ancestry_was) with the new one
path.

This addresses #33

### DIFF
--- a/test/concerns/update_test.rb
+++ b/test/concerns/update_test.rb
@@ -1,0 +1,19 @@
+require_relative '../environment'
+
+class UpdateTest < ActiveSupport::TestCase
+  def test_node_creation_in_after_commit
+    AncestryTestDatabase.with_model do |model|
+      children=[]
+      model.instance_eval do
+        attr_accessor :idx
+        self.after_commit do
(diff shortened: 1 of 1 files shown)


kbrock:
a) this works in active record 5.2
b) this works in after commit hook
c) this method is changing a bunch when going across to using ids (instead of the string munging)

If you need for AR < 5.2, please let me know

this method is being transformed

kbrock:
want to introduce test - the bug is in the bad implementtaion of the code that asks "what is the path?"
which munged was and current

PR #264, opened 2016-02-09
Fix various issues where tests were not running

Fryguy:
- Fix issue where "minitest/autorun" must be required for newer versions
  of Rails
- Made sqlite3 the default local database when not using appraisals
- Handle non-installed databases when running `appraisal rake test`
  locally


Fryguy:
@kbrock Please review.

Fryguy:
Supercedes #251

PR #265, opened 2016-02-11
Adds support for primary_key_format as lined out in the README

walski:
Should fix https://github.com/stefankroes/ancestry/issues/221 and https://github.com/stefankroes/ancestry/issues/241

I understand that there is https://github.com/stefankroes/ancestry/pull/262 which affects this PR as well. I'm fine with fixing support for this as soon as #262 is merged.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -119,6 +119,9 @@ The has_ancestry methods supports the following options:
   :depth_cache_column    Pass in a symbol to store depth cache in a different column
   :primary_key_format    Supply a regular expression that matches the format of your primary key.
                          By default, primary keys only match integers ([0-9]+).
+                         Make sure to use an expression that matches one but also multiple IDs,
+                         separated by a / (slash). The complete default expression is:
+                         /\A[0-9]+(\/[0-9]+)*\Z/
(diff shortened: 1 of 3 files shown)


kbrock:
Hi @walski Maybe #262 is trying to do too much. ideas how to pair that down?
Also, iof #262 is merged, would that meet your needs as well?

walski:
@kbrock #262 would totally work for me. I need to support UUIDs in the ancestry column, that's all.

I do not completely understand what holds #262 back from being merged or what is the big problem there. Happy to help, though if I can.

kbrock:
@walski That was my interpretation of half a dozen PR/issues.

I was hoping to get some feedback from @stefankroes before merging it. Or at least a thorough review pointing out if some of my assumption are ok. I'll comment over there.

kbrock:
@walski just read over your comments again.

I hadn't realized that `:primary_key_format` was added to the docs, but looks like it was lost somewhere.

Updated #262 - seems our two are converging. Although, I'm specifying a single symbol vs including the delimiter (Which is the original usage)

kbrock:
thanks for the help here.

I'm going to remove this feature.
let people setup their own validations if they find it necessary.

PR #268, opened 2016-03-04
For arranged, use index instead of separate hash

kbrock:
`arranged` is basically just `index[parent_id]`

So store the parent_id and don't have a separate `arranged` hash

/cc @Fryguy does this make sense to you?

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -42,23 +42,20 @@ module Ancestry
     # Arrange array of nodes into a nested hash of the form
     # {node => children}, where children = {} if the node has no children
     def arrange_nodes(nodes)
-      arranged = ActiveSupport::OrderedHash.new
+      root_id = nil
       min_depth = Float::INFINITY
       index = Hash.new { |h, k| h[k] = ActiveSupport::OrderedHash.new }
 
       nodes.each do |node|
-        children = index[node.id]
-        index[node.parent_id][node] = children
-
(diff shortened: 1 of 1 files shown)


lib/ancestry/class_methods.rb
```diff
@@ -42,23 +42,20 @@ def arrange options = {}
...
-      arranged = ActiveSupport::OrderedHash.new
+      root_id = nil
       min_depth = Float::INFINITY
       index = Hash.new { |h, k| h[k] = ActiveSupport::OrderedHash.new }
 
       nodes.each do |node|
-        children = index[node.id]
-        index[node.parent_id][node] = children
-
+        index[node.parent_id][node] = index[node.id]
```
Fryguy: I had this separated intentionally to make the code easier to read...the children variable here matched the comment up above.  I'm ok with removing it, but wasn't sure if you had other reasons other than removing that intermediate variable.

Fryguy:
After reviewing, I believe the reason I had this was for multiple root nodes, and this would break that.  I don't think we have a test for it, so I'm wondering if we can add a simple test for that first.  Even a degenerate test with 2 roots nodes and nothing else should be sufficient.

kbrock:
@Fryguy Yes, having depth makes this algorithm work very well here.
Think I remember having a way to get this confused, but need to track it down.

I was playing with this idea because in miq, I needed a function like this without the `depth` calculation.

PR #271, opened 2016-03-08
Leave Scope Added

ImDineshSaini:
Leave scope added and now the function can be used as ModelName.leaves

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -16,7 +16,7 @@ module Ancestry
         end
       end
     end
-    
+
     # Scope that returns all the leaves
     def leaves
       id_column = "#{table_name}.id"
@@ -27,7 +27,7 @@ module Ancestry
         group(id_column).
         having('COUNT(c.id) = 0')
     end
-    
+
     # Orphan strategy writer
     def orphan_strategy= orphan_strategy
       # Check value of orphan strategy, only rootify, adopt, restrict or destroy is allowed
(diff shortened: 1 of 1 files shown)


ImDineshSaini:
Leave option added - dependent function added

antstorm:
@ImDineshSaini this PR is a total rip off of #246

ImDineshSaini:
Yes and one more changes required I have seen the Rails 4 is work with mysql2 gem so  text_type = if ActiveRecord::Base.connection.adapter_name.downcase == 'mysql2' replaced by  text_type = if ActiveRecord::Base.connection.adapter_name.downcase == 'mysql'

ImDineshSaini:
this PR is a total rip off of #246  + Fix the mysql connection error with rails 4 and utf issue.

antstorm:
@ImDineshSaini all of that is fixed in #246 and tested

ImDineshSaini:
I have tries with Rails 4.2.3 and mysql and it give me error and reason is that mysql2 is use as connection not mysql.

lib/ancestry/class_methods.rb
```diff
@@ -17,6 +17,17 @@ def scope_depth depth_options, depth
       end
     end
 
+    # Scope that returns all the leaves
+    def leaves
+      id_column = "#{table_name}.id"
+      id_column_as_text = sql_cast_as_text(id_column)
+      parent_ancestry = sql_concat("#{table_name}.#{ancestry_column}", "'/'", id_column_as_text)
+
+      joins("LEFT JOIN #{table_name} AS c ON c.#{ancestry_column} = #{id_column_as_text} OR c.#{ancestry_column} = #{parent_ancestry}").
```
kbrock: is there a way to use an existing scope here?
Only adding the `group` and `having`

Also, should `id` be `primary_key`?

kbrock: better yet, is there a way to extract this concept so we can use it more?

lib/ancestry/class_methods.rb
```diff
@@ -202,5 +208,26 @@ def rebuild_depth_cache!
...
+
+    def sql_cast_as_text column
+      #need mysql2 as new rails version support the mysql2 gem only
+      text_type = if ActiveRecord::Base.connection.adapter_name.downcase == 'mysql2'
+        'CHAR'
+      else
+        'TEXT'
+      end
+      # updated mysql thrown error if utf is not defined
+      "CAST(#{column} AS #{text_type}) collate utf8_unicode_ci"
```
kbrock: think the `collate` is better up with mysql only

antstorm: @kbrock why are you even looking at this PR? The original one is #246

kbrock: please be nice

kbrock:
Closing in favor of #246

PR #272, opened 2016-03-08
Unscope ancestors before touching them.

samtgarson:
I've been having an issue whereby creating a child using the scoped syntax, (e.g. `grandchild.children.create`) will not touch the ancestors due to the implicit scope applied.

It looks like back when the `unscoped` method was applied to various operations in eeadeef and cd268b6, the `touch` stuff wasn't implemented yet.

Test created to demonstrate the use case.

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -76,7 +76,7 @@ module Ancestry
         if self.ancestry_base_class.touch_ancestors
 
           # Touch each of the old *and* new ancestors
-          self.class.where(id: (ancestor_ids + ancestor_ids_was).uniq).each do |ancestor|
+          unscoped_current_and_previous_ancestors.each do |ancestor|
             ancestor.without_ancestry_callbacks do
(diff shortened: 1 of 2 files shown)


lib/ancestry/instance_methods.rb
```diff
@@ -76,7 +76,7 @@ def touch_ancestors_callback
         if self.ancestry_base_class.touch_ancestors
 
           # Touch each of the old *and* new ancestors
-          self.class.where(id: (ancestor_ids + ancestor_ids_was).uniq).each do |ancestor|
+          unscoped_current_and_previous_ancestors.each do |ancestor|
```
kbrock: Maybe:

```ruby
ancestry_base_class.unscoped.where(id: (ancestor_ids + ancestor_ids_was)).each do |ancestor|
```

Don't think the `uniq` or the `self` is needed, but I haven't researched.

Think the separate method may be a bit much unless you see it used elsewhere.

Either way, I'd prefer the single line version

PR #277, opened 2016-05-10
Fix child_ancestry for new models

Bogadon:
The child_ancestry method was returning id if the ancestry column was previously blank, for example for a new model. 

This is incorrect for non-root nodes, and leads to broken ancestry relations. I discovered this due to a model mis-creating children during an after_create callback. 

This PR will fix this, but may break whatever issue the former condition was added to address.

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -90,7 +90,7 @@ module Ancestry
       # New records cannot have children
       raise Ancestry::AncestryException.new('No child ancestry for new record. Save record before performing tree operations.') if new_record?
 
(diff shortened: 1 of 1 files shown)


kbrock:
Thanks.

Could you add a test to show the issue?
Think the PR is very close

kbrock:
@Bogadon Looks like the original code was that way for a reason.
Could you post a test and I could look into a fix that wouldn't break the tests?

kbrock:
bouncing to trigger travis again on this one

kbrock:
@Bogadon I put together code that I think fixes this, but without a test case, I do not know for sure.

Could you provide a test case, or describe what you are doing that you have a non-created node with children.

Thanks

kbrock:
Closing this for now.

I need more to be able to fix this.
Please post a comment with a test (or a ruby snippet) to show how this can be reproduced.

Thanks

PR #284, opened 2016-08-11
Update README.rdoc

vasinov:
Updated some badges to svg and added a Hakiri security badge.


kbrock:
thank you.  sorry for the delay

PR #285, opened 2016-09-17
Add siblings-only finders (excluding the targeted node)

chikamichi:
Hi,

This PR adds support for siblings-only finders and scope.

Excluding the targeted node from the list of its siblings is as simple as `node.siblings.reject { |s| s == node }`, but it might make more sense to handle that within ancestry itself so as to leverage a cleaner SQL query and improve consistency (eg. `Model.siblings_only_of(node)`, `Model#siblings_only_ids`).

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -76,28 +76,30 @@ You can also create children through the children relation on a node:
 
 To navigate an Ancestry model, use the following methods on any instance / record:
 
-  parent           Returns the parent of the record, nil for a root node
-  parent_id        Returns the id of the parent of the record, nil for a root node
-  root             Returns the root of the tree the record is in, self for a root node
-  root_id          Returns the id of the root of the tree the record is in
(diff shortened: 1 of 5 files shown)


chikamichi:
Specs actually are passing, something's wrong on Travis for Ruby 1.9.3 when it comes to the config and some gems cannot be installed:

> An error occurred while installing {foobar} (x.y.z), and Bundler cannot continue.

kbrock:
@d-m-u wondering
- does it make sense to make siblings a configuration setting?
- is there an agreed upon standard for what you call all my parent's children?
- I think I like the simpler query and count > 1 vs complex query > 0

kbrock:
related to #488

kbrock:
I'm thinking this should be a breaking change and a config parameter would switch between the two?

thinking out loud and would love any feedback.

Issue #283, opened 2016-08-10
Touch and STI incorrect constraint on .type


brendon:
I've been having a play with the `touch: true` feature in combination with a tree that uses `STI`. Unfortunately the way the touch callback works, Rails places a constraint on the type column that matches the type of the original saved node when selecting ancestors. This obviously fails if a parent node is of a different type.

I've patched the method in my app and it seems to work. I'm not sure what would happen should the type column not exist though but it's a start:

```
# Touch each of this record's ancestors
  def touch_ancestors_callback

    # Skip this if callbacks are disabled
    unless ancestry_callbacks_disabled?

      # Only touch if the option is enabled
      if self.ancestry_base_class.touch_ancestors

        # Touch each of the old *and* new ancestors
        self.class.unscope(where: :type).where(id: (ancestor_ids + ancestor_ids_was).uniq).each do |ancestor|
          ancestor.without_ancestry_callbacks do
            ancestor.touch
          end
        end
      end
    end
  end
```

brendon:
Well that's hilarious! I was looking through the current code to see if it was still an issue and found `unscoped_where` and thought that it sounded familiar. Turns out I [added that a while back](https://github.com/stefankroes/ancestry/pull/348/files#diff-46e331d157085e6c560dd10be25071dcR327) lol! By the looks of things, this shouldn't be an issue anymore because we unscope the whole `where` clause which should include the `type` column. Thanks for following this up :)

brendon:
What's worse is I removed the patch code from my project when I upgraded ancestry and didn't even think to ping this issue. Sorry ☺️

PR #287, opened 2016-09-27
Handle STI

splybon:
touch: true does not work when used with STI.

Fixes: https://github.com/stefankroes/ancestry/issues/283


kbrock:
thanks @splybon - Surprised I didn't run into this in the past.

1. would a generalized `unscoped` make more sense? - honest question rather than a polite request. Thinking we'd avoid default scopes (which may be good or bad - not sure) Then again, the current tests (non STI) seem to be working.
2. Could you add a test case or at least give a trimmed down model from your code so I can reproduce the problem in a test? Need the test so we don't break it in the future by mistake. (and across rails versions)

thanks for the PR

splybon:
I don't think a generalized unscoped is totally necessary.  There may be default scopes that people want on there.

I'll add a test to the PR

kbrock:
I fixed the build. so if you push again - it should go green for you

kbrock:
kicking travis

kbrock:
@splybon thanks for the tests and this fix. I'm going to push out a release of ancestry without this change. then I'm going to re-evaluate dropping active record 3.0 support for ancestry 3.0. Would prefer to support as many versions as possible on master

kbrock:
running again with different gem requirements. (think it may still fail for 3.2)

kbrock:
ok, looks like `unscoped` is not yet defined in rails 3.2 (but this is working for all newer versions)

lib/ancestry/instance_methods.rb
```diff
@@ -76,7 +76,7 @@ def touch_ancestors_callback
         if self.ancestry_base_class.touch_ancestors
 
           # Touch each of the old *and* new ancestors
-          self.class.where(id: (ancestor_ids + ancestor_ids_was).uniq).each do |ancestor|
+          self.class.unscope(where: :type).where(id: (ancestor_ids + ancestor_ids_was).uniq).each do |ancestor|
```
kbrock: Would it make sense to use `ancestry_base_class` here instead?

Also, could you add a test so we can be sure this doesn't break in the future?

Thanks

kbrock:
We've dropped 3.x support
but I think we have an unscoped for this.

Does master fix this for you?

kbrock:
@splybon There have been many fixes around `unscope`

Could you share a test or at least try `master` with your code and see if it is already fixed?

Thanks

PR #291, opened 2016-10-20
Consistently access ancestry key columns

kbrock:
We access these columns a few different ways.
Use `get_ancestry_column` and `get_primary_key_column` in all places


kbrock:
opting to delete these rather than make them better

PR #294, opened 2016-10-21
Use case sensitive 'like' for rails 5

kbrock:
Not my favorite code fix.
Have longer term plans to improve this fix. But this
is the least invasive. It can make the queries up to
10x faster


Fryguy:
LGTM :+1: ... was this tested against the new Rails 5 tests?

PR #300, opened 2016-10-28
Don't require database connection at class loading

### LINKED ISSUE #299: v2.2.1 is unusable in gems with models
I'm using ancestry in one of my gems (see [ledermann/keepr](https://github.com/ledermann/keepr)). There are models in this gem using `has_ancestry` in their class definition (e.g. [here](https://github.com/ledermann/keepr/blob/master/lib/keepr/group.rb))

After upgrading to v.2.2.1 this is not possible anymore because auf #259. I fixed it for running the tests of the gem, but using my gem in the application fails.

This is a load order issue: Accessing `connection.adapter_name` (made at class level in v2.2.1) requires a database connection to be established. But in a standard Rails application, models in gems are loaded **before** the database is connected. This results in the following error message on application startup:

    Gem Load Error is: No connection pool with id primary found.

Any ideas?

ledermann:
Check the database adapter **inside** the lambda, so it gets not executed at class loading time anymore.

Fixes #299


kbrock:
Thank you @ledermann - Darn. That was a dumb mistake.

PR #306, opened 2016-11-01
cache travis bundle

kbrock:
Since all the builds use a different version of the gems, not sure if bundle cache will help here

testing it out


kbrock:
didnt seem to buy us much

PR #311, opened 2016-12-09
Simplify ActiveRecord version checks

Fryguy:
@kbrock Please review.

### DIFF
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -44,16 +44,14 @@ class << ActiveRecord::Base
     scope :subtree_of, lambda { |object| where(to_node(object).subtree_conditions) }
     scope :siblings_of, lambda { |object| where(to_node(object).sibling_conditions) }
     scope :ordered_by_ancestry, lambda {
-      if %w(mysql mysql2 sqlite postgresql).include?(connection.adapter_name.downcase) &&
-        defined?(ActiveRecord.version) && ActiveRecord.version.to_s >= "5"
+      if %w(mysql mysql2 sqlite postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::MAJOR >= 5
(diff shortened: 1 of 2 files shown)


lib/ancestry/has_ancestry.rb
```diff
@@ -44,16 +44,14 @@ def has_ancestry options = {}
     scope :subtree_of, lambda { |object| where(to_node(object).subtree_conditions) }
     scope :siblings_of, lambda { |object| where(to_node(object).sibling_conditions) }
     scope :ordered_by_ancestry, lambda {
-      if %w(mysql mysql2 sqlite postgresql).include?(connection.adapter_name.downcase) &&
-        defined?(ActiveRecord.version) && ActiveRecord.version.to_s >= "5"
+      if %w(mysql mysql2 sqlite postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::MAJOR >= 5
```
Fryguy: I was thinking of moving this logic to a class method we can memoize and that describes *why* we are doing this.  From reading the next line it seems it's to support the `coalesce` operator?  Is that true, and if so, why the need for the Rails 5 check?

kbrock: We check rails 5, but technically it is the arel version that comes with rails 5.

Huh. we should only need to check this for the extra parameter in `matches`. Odd it ended up here.

lib/ancestry/instance_methods.rb
```diff
@@ -257,7 +257,7 @@ def sibling_of?(node)
     def descendant_conditions
       t = get_arel_table
       # rails has case sensitive matching.
-      if defined?(ActiveRecord.version) && ActiveRecord.version.to_s >= "5"
+      if ActiveRecord::VERSION::MAJOR >= 5
```
Fryguy: I was thinking the same thing, but I didn't want to change the *other* two places to class definition if statements, so I let it be.  I'll leave it up to you how to proceed after this PR.

PR #314, opened 2016-12-18
Use pluck instead of select + map

njakobsen:
This avoids instantiating records with only a subset of the expected columns, which may have unintended consequences.

Specifically, our project uses default_value_for, which sets default values when the record is instantiated. When the `_ids` methods use `select`, records are instantiated without all the expected columns, causing an ActiveRecord exception when attempting to initialize the missing columns with default values. This change avoids the need to instantiate records at all, instead using the `pluck` method to return the primary key column values.

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -206,7 +206,7 @@ module Ancestry
     end
 
     def child_ids
-      children.select(self.ancestry_base_class.primary_key).map(&self.ancestry_base_class.primary_key.to_sym)
+      children.pluck(self.ancestry_base_class.primary_key)
     end
 
     def has_children?
@@ -235,7 +235,7 @@ module Ancestry
     end
 
     def sibling_ids
-      siblings.select(self.ancestry_base_class.primary_key).collect(&self.ancestry_base_class.primary_key.to_sym)
+      siblings.pluck(self.ancestry_base_class.primary_key)
     end
 
     def has_siblings?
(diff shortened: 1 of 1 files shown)


njakobsen:
Travis failure appears to test suite config problem, and be unrelated to the changes in the PR.

kbrock:
Aah, I get it. Active Record 3.x does not support `pluck`.

Frustrating, since this change is so much better than the non-pluck version, this may push us towards dropping rails 3.x.

njakobsen:
This could be refactored to use `ActiveRecord::Base.connection.select_values` to keep AR 3.x support.

njakobsen:
Hmm, `select_values` doesn't typecast depending on the database adapter.  Unless there's a simple way to typecast the value automatically, since we're using `primary_key`, I hesitate to simply `to_i` all the values returned since we don't know that they are going to be integers. Thoughts?

kbrock:
Hmm.  Going with `pluck` simplifies the statement. Going with `select_values` just makes the thing more convoluted. my preference is

1. keep select / map
2. drop AR 3.0
3. introduce better sql detection (runs into detection madness - but we kinda want this for other stuff too)

sorry, thinking on this.

chrisarcand:
@kbrock `pluck` is in AR 3.2.1+

Lots of people are still on Rails 3, but no one should be on 3.0.20 (the version the test suite is running). 

The test suite should be updated to handle latest and multiple Rails releases, then this PR can be rerun and reviewed with the `pluck` present. I can assist with if needed.

njakobsen:
Force pushed back to the pluck commit

kbrock:
triggering a new build

kbrock:
@njakobsen per [old travis failure](https://travis-ci.org/stefankroes/ancestry/jobs/185034511) I'm thinking that sqlite's any datatype will do style will warrant changing tests.

I also wonder if `string`s instead of `int`s coming through will screw up devs.
(hope no one is running sqlite in production)

njakobsen:
I would hesitate to change the return value, even if it is simply int to string. I'm sure someone is comparing a list of ids, which will suddenly fail. `pluck` casts return values using column information, so we should be good.

PR #324, opened 2017-03-27
adds ability to explicitly define child classes when using STI.

pratuat:


kbrock:
Hello @pratuat looks like it works for all but 3.2 (not a big thing)

Could you explain the reasoning behind this request?
I'm sure it make sense, I just can't figure out the use case.

If the base class is the parent of all the children, this seems to work well for STI.
Not sure why this would fail.

But again, I want to understand more so we can get something like this in

pratuat:
Hello @kbrock 

Only thing I was trying to achieve was to set STI type appropriately while creating children without the need for being explicit in build params.

For instance, while using ancestry tree in a **Organization** model, the model can have three sub-classes **Importer**, **Distributor** and **Dealer**. **Importer** can only have  **Distributor** children and **Distributor**  can only have **Dealer**.

So I just wanted `@organization.children.build` to spit out appropriate sub_class instance without the need of being explicit as `@organization.children.build(type: return_child_klass_of(@organization))`.

kbrock:
Thanks, I need to push around a few classes and see how this falls out.

Ping me again in a week if this falls through the cracks

kbrock:
need some tests before I merge it.
Having said that, still a little on the fence.
Also want to kick to see if it will pass tests.

kbrock:
I'm thinking overriding `children` would be the best way to implement this feature.

This feature seems like an edge case. Concerned about maintaining this code

kbrock:
I went ahead and resolved the conflict and rebased
there are no tests here, so I am not sure that this code works.

I'm going to close this for now.

This change forces `self` to be in memory to find `children`.
In trying to remove all the N+1 and object allocation, `children` can be determined from an instance method or list of objects/integers.

So handling this case will force us to load these objects into memory and/or introduce complex code if the source object classes vary.

I am wondering about all the `ancestry_base_class` calls. It is verbose and has the potential to be an issue with great concepts like this PR introduces.

Sorry to close

PR #329, opened 2017-04-27
Restore compatibility with `acts_as_tree`

### LINKED ISSUE #103: acts_as_tree alias will still be defined if a colliding gem name occurs lexically after ancestry
For instance if I install the closure_tree gem the ancestry gem will still define acts_as_tree even though it is defined in closure_tree as well.  This breaks the closure_tree gem definition of acts_as_tree.  My guess is that ancestry is loaded before closure_tree and that is why the code breaks.

crazymykl:
Fixes #103 

It may make sense to make these modules to mix in instead of monkeypatching, but I'm just fixing this one small bugaboo.

### DIFF
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -97,9 +97,13 @@ class << ActiveRecord::Base
 end
 
 ActiveSupport.on_load :active_record do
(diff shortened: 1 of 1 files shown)


kbrock:
@crazymykl thanks

PR #340, opened 2017-05-10
Fix check for new parent_id in #parent_id=

domcleal:
a8e95fa mistakenly replaced a nil check for the new parent ID in the
parent_id setter with a check for `ancestry?`. The setter would then
only set the parent if the model already had ancestors, rather than if
a parent ID had been passed.


kbrock:
Thanks dom

PR #341, opened 2017-05-10
Add leaves instance method

llenodo:
Add an instance method `leaves` which returns all leaf nodes from a given node.

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -279,6 +279,12 @@ module Ancestry
       subtree(depth_options).pluck(self.ancestry_base_class.primary_key)
     end
 
+    # Leaves
+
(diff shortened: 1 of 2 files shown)


kbrock:
Currently working towards getting all these methods using scopes instead of moving all records through ruby. Want to hold off on this for just a little bit

kbrock:
@llenodo any ideas how to do this in sql and not ruby?

kbrock:
Also, I think this is going to generate way too many sql calls.

llenodo:
Might be a way to do w/ SQL but would be far more complicated. IMO better to put this in now and refactor later but if you dont like the implementation feel free to reject this PR and create an open issue instead

kbrock:
see also #246

PR #344, opened 2017-06-01
Added descriptive alias :has_parent? to existing :ancestors? method

stowersjoshua:
I had assumed there was already a `:has_parent?` method, but was surprised to get an error when I tried.
There is an undocumented `:ancestors?` method which checks for a value in the parent field, which is what I was hoping for from a `:has_parent?` method.
This PR aliases the two methods and updates the documentation to describe the alias _only_.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,7 @@ record:
     path             Scopes model on path records of the record
     children         Scopes the model on children of the record
     child_ids        Returns a list of child ids
+    has_parent?     Returns true if the record has a parent, false otherwise
     has_children?    Returns true if the record has any children, false otherwise
     is_childless?    Returns true is the record has no children, false otherwise
     siblings         Scopes the model on siblings of the record, the record itself is included*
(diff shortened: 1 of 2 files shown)


kbrock:
I'm sorry that `ancestors?` did not get into the docs. Will need to add that.

So, there is `parent_id?` and `ancestors?`. Do either of those names work for you?
(I suppose `!is_root?` is probably the same)

Not a fan of the naming convention `has_*?` - I prefer names that are of the form `#{relation}?` Rails introduced the concept of `#{relation}? == #{relation}.present?`

Do you feel the name `parents?` is better than `ancestors?`

stowersjoshua:
I opted for the `has_#{relation}?` method both because I prefer how it reads, and to continue with the convention I saw set by `has_children?` and `has_siblings?`.
Perhaps it is worth adding both `parents?` and `has_parents?`, in order to let individual developers that are using Ancestry stick with whichever convention they already follow.

To answer your last question, I think `parents?` probably would work better than `ancestors?`.

PR #345, opened 2017-06-05
Remove duplicate scope :path_of

ledermann:
This avoids warnings like this:

    Creating scope :path_of. Overwriting existing method MyModel.path_of.


kbrock:
thanks

PR #348, opened 2017-07-03
Be more surgical about unscoping; only unscope :where

brendon:
Related to: https://github.com/swanandp/acts_as_list/pull/265

I’ve been running into deadlocks in my app. In the past I’ve identified
these being caused by `acts_as_list` being too greedy in its unscoping
(removing default order scopes, and thus creating situations where the
table is searched from opposite ends resulting in a deadlock).

Ancestry seems to be doing the same thing. I’ve made a modification
that ensures we only remove the `:where` scope, leaving other scopes
(like `:order`) intact. The tests still pass (locally). I’ve added in a
workaround for Rails 3.2 that doesn’t support this functionality.

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -99,9 +99,9 @@ module Ancestry
       parents = {}
       exceptions = [] if options[:report] == :list
 
-      self.ancestry_base_class.unscoped do
+      unscoped_where do |scope|
         # For each node ...
-        self.ancestry_base_class.find_each do |node|
+        scope.find_each do |node|
           begin
             # ... check validity of ancestry column
             if !node.valid? and !node.errors[node.class.ancestry_column].blank?
@@ -137,9 +137,9 @@ module Ancestry
       parents = {}
       # Wrap the whole thing in a transaction ...
       self.ancestry_base_class.transaction do
-        self.ancestry_base_class.unscoped do
+        unscoped_where do |scope|
           # For each node ...
-          self.ancestry_base_class.find_each do |node|
+          scope.find_each do |node|
             # ... set its ancestry to nil if invalid
             if !node.valid? and !node.errors[node.class.ancestry_column].blank?
               node.without_ancestry_callbacks do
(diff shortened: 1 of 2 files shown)


brendon:
Thanks @kbrock :)

brendon:
Hi @kbrock, just a quick followup. I was wondering when you're planning on releasing the next version of the gem? That'll help me decide wether to bundle in the master branch or hold off for a bit :)

kbrock:
@brendon for some reason, the build is failing for me on the local machine (when master before this merge works). need to dig into this more

brendon:
Thanks @kbrock, ah yes RE the Rails 3.2 and Mysql 5.7 bug. I remember fixing that in one of my projects.

Well done on still supporting 3.2 by the way. I'm now thankfully off it, but I know many projects are still on it as it does take a while to upgrade big projects!

kbrock:
ok, I rebased locally.

```
DB=sqlite3 BUNDLE_GEMFILE=gemfiles/gemfile_60.gemfile bundle exec rake

DEPRECATION WARNING:
Class level methods will no longer inherit scoping from `create!` in Rails 6.1.
To continue using the scoped relation, pass it into the block directly.
To instead access the full set of models, as Rails 6.1 will, use
`AncestryTestDatabase::TestNodeCreationInAfterCommitTestNode.unscoped`.
(called from children_of at lib/ancestry/materialized_path.rb:34)
```

Let me reiterate, I'm worried about removing all scopes, especially the order.
Also, I'm not sold that this fix will fix things.

I'm pretty sure I tried passing in scopes as suggested and the code base broke in odd ways.
So please respect that clicking merge may introduce deadlocks for some use cases and it will probably not solve this deprecation warning.

brendon:
Hi @kbrock, was that post a mistake? I got it via email but I think you must have deleted it? :)

kbrock:
hi @brendon yes.

There has been a lot of discussions in #442 about changing this commit to remove a deprecation warning. But I don't believe this is the culprit.

Sorry to send you a message and have a great week

brendon:
That's all good. I've linked into an `acts_as_list` issue that documents a bit of us trying to sort it out. We seem to have stalled though :)

PR #356, opened 2017-08-09
Fix order_by_ancestry_and

### LINKED ISSUE #353: [Regression] ordered_by_ancestry_and broken
Hi,

I just tried to update ancestry 2.2.2 -> 3.0.1 and our test suite froze in an infinite recursion. I downgraded to 3.0.0 and the problem persists.

Digging in, the problem appear when `ordered_by_ancestry_and` is called. As it is been [redefined](https://github.com/stefankroes/ancestry/compare/v2.2.2...v3.0.0#diff-8f2e9f299de2ed3dfab16cceb84abf43R55) it is a scope calling itself which matches the observation. I assume, it's meant to call `ordered_by_ancestry` which now accepts additional custom ordering.

kbrock:
fixes #353


kbrock:
This fixes the bug, but it is apparent that sorting is currently not working for most postgres implementations (c locale works)

PR #357, opened 2017-08-22
Directly generate the index in the migration

tilsammans:
Rails has built-in syntax to add a column plus associated index.

This means we can drop a bit of instruction. It also guarantees consistency.


kbrock:
@tilsammans Thanks

PR #359, opened 2017-08-30
Fix warnings in Rails 5.1 about using dirty tracking

westonganger:


macfanatic:
Any hints as to why this wasn't merged @westonganger?

westonganger:
I remember creating this PR then sometime after thinking that possibly this library was not the cause. But honestly I just sort of learned to live with the Rails deprecations which however are extremely annoying.

If someone else can verify this code, this PR could be looked into again.

macfanatic:
Thanks for the info.  I was able to identify this as the culprit of a ton of deprecation warnings in our codebase (noticed in rspec test runs) with upgrade to Rails 5.1

I had implemented something similar but didn't have any luck keeping the tests passing locally so tabling for now.

Issue #362, opened 2017-10-05
.arrange method filters out some records.


trafium:
Hi, I am implementing a forum using ancestry gem and I am having some issues with how .arrange method handles some of records. Imagine I have queried data like this:
```
[#<Foo id: 1, ancestry: nil>,
#<Foo id: 2, ancestry: "1">,
#<Foo id: 4, ancestry: "3">]
```
So i have one root node, one child of that node and one child of root which is not present in this particular query result. I expected the result of .arrange method to be something like
 ```
{
  #<Foo id: 1, ancestry: nil> => {
    #<Foo id: 2, ancestry: "1"> => {}
  },
  #<Foo id: 4, ancestry: "3"> => {}
}
```
but instead  it completely filters out Foo#4, because input data doesn't apparently contain it's parent. I understand that there could be reasons for this behaviour (like keeping same hash depth and record ancestry depth), but shouldn't the arrange method just arrange, not filter?
My use case is that I am fetching all users's comments (independent of ancestry depth) and all their children to show user answer branches  for his comments.

**Comment by kbrock** [comment]
thanks @trafium 

@trafium - what database are you using? Are the values you are showing actual, or an approximation?

@Fryguy wonder if new arrange tweaked something (especially on postgres that uses an odd order)

**Comment by trafium** [comment]
@kbrock thank you for quick response

Values are approximation indeed, but I tested it with exactly 3 records with matching ancestries.

I actually spent some time to analyse code and figured out that this is happening not exactly because of ordering in arrange method itself. All records successfully get ordered (basically by ancestry depth first) and passed to arrange_nodes method. Then in arrange method min_depth is set by first node's depth and node is assigned to resulting hash directly only if it's depth matches min_depth. 
So basically Foo with id: 4 never gets assigned as key to resulting array directly neither being assigned as some other record's child because it's parent is missing from input nodes.

The solution I used for my personal case is following:

```
def custom_arrange options = {}
  # Ancestry order removed because with my use case even when I manage to arrange 
  # all nodes, custom order will not work properly because node with depth 0 will always 
  # precede node with depth 1, also it allows ordering before invoking arrange
  arrange_nodes self.where(options)
end

def custom_arrange_nodes(nodes)
  nodes_ids = nodes.map(&:id)
  arranged = ActiveSupport::OrderedHash.new

  # assigning min_depth right from the start to work with nodes not being ordered by ancestry
  min_depth = nodes.map(&:depth).min

[...]

PR #363, opened 2017-10-10
Fix arrange/arrange_nodes methods filtering out some nodes.

trafium:
https://github.com/stefankroes/ancestry/issues/362

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -29,27 +29,26 @@ module Ancestry
 
     # Arrangement
     def arrange options = {}
-      # Get all nodes ordered by ancestry and start sorting them into an empty hash
-      arrange_nodes self.ancestry_base_class.reorder(options.delete(:order)).where(options)
+      # Get all nodes and start sorting them into an empty hash
+      arrange_nodes self.where(options)
     end
 
     # Arrange array of nodes into a nested hash of the form
     # {node => children}, where children = {} if the node has no children
     def arrange_nodes(nodes)
-      arranged = ActiveSupport::OrderedHash.new
-      min_depth = Float::INFINITY
-      index = Hash.new { |h, k| h[k] = ActiveSupport::OrderedHash.new }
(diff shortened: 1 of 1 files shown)


vzharkov:
@trafium this PR break the ordering for the `arrange` method.

lib/ancestry/class_methods.rb
```diff
@@ -29,27 +29,25 @@ def orphan_strategy= orphan_strategy
...
+      arrange_nodes self.where(options)
     end
 
     # Arrange array of nodes into a nested hash of the form
     # {node => children}, where children = {} if the node has no children
     def arrange_nodes(nodes)
-      arranged = ActiveSupport::OrderedHash.new
-      min_depth = Float::INFINITY
-      index = Hash.new { |h, k| h[k] = ActiveSupport::OrderedHash.new }
+      arranged  = ActiveSupport::OrderedHash.new
```
kbrock: hashes are ordered by default. can we just use a regular hash?

lib/ancestry/class_methods.rb
```diff
@@ -29,27 +29,25 @@ def orphan_strategy= orphan_strategy
 
     # Arrangement
     def arrange options = {}
-      # Get all nodes ordered by ancestry and start sorting them into an empty hash
-      arrange_nodes self.ancestry_base_class.reorder(options.delete(:order)).where(options)
```
kbrock: not sure how removing the ordering helps the pr

The issue with removing this ordering is the tree will still get created, but the nodes will not be in the secondary desired order.

Also of note, the ancestry column does not have `c` collation. So I do not trust sql's `order` right now.

kbrock:
Thank you for opening this PR.

future suggestion: make a local branch rather than creating this on your master.

I cant merge without a test, or with it breaking existing tests.
I also don't know how we are to figure out how to handle the missing 

Please reopen / ping me if you want to work through this

kbrock:
@trafium so sorry for closing this.

I pulled this change (a little tweaked) into #415

PR #364, opened 2017-10-12
Adding scope for leaves

fursich:
### what it does

This PR is to adding scopes (with several related instance methods) for leaves.
Looks like there have been quite a few requests around this feature, so hoping it would give some solutions.

### thoughts..

As you can see the basic approach here consists of two steps:

1) collect all ancestor_ids from the model in order to later identify childless nodes (i.e. leaves)
2) using the ids above, search childless nodes that belongs to the subtree (incl. descendants and itself) of the given node. 

Another way is to use LEFT JOIN to its own table, eg.

```sql
  SELECT DISTINCT *
     FROM categories AS c
     LEFT OUTER JOIN categories AS c2
     ON ( CONCAT(c.ancestry, '/', c.id) = c2.ancestry) OR ( c.id = c2.ancestry)
    WHERE c2.id is null;
```

I actually implemented it with complex Arel, but that requires a slight hustle to handle database-dependent function `CONCAT` (`||` for sqlite3, or `+` in some cases).

The good point is it only creates single SQL query vs two in the earlier example - however I chose earlier approach as it seemed much cleaner codes imho.

Hope that makes sense! (and thanks as always for maintaining the awesome gem!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,9 @@ record:
     is_only_child?   Returns true if the record is the only child of its parent
     descendants      Scopes the model on direct and indirect children of the record
     descendant_ids   Returns a list of a descendant ids
+    leaves           Scopes the model on childless nodes from among the record's subtree
+    leaf_ids         Returns a list of leaf ids
+    leaf? is_leaf?   Returns true if the record is a leaf node (ie. childless), false otherwise
     subtree          Scopes the model on descendants and itself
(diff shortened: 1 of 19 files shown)


fursich:
Hmm.. `bundle install` seems to be constantly failing with ruby 1.9.3.

Sorry for the current branch getting confusing with irrelevant codes, let me re-submit the pull request once again without changes on gemfiles.

PR #370, opened 2017-12-26
 Replace deprecated attribute_was

Bramjetten:
The behavior of `attribute_was` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `attribute_before_last_save` instead.


kbrock:
Thank you for explaining the current issue.

I've changed over the code to use `was` (bad) to get everything passing in 5.2
I'm currently going through and trying to consolidate the accesses to the `was` variables.

Please let me know if you have any suggestions for what we now do on master.

kbrock:
Thanks for the PR.

Think this needed a few more changes
Let me know if #387 meets your needs

Thanks

PR #373, opened 2018-01-05
Fix "Dangerous query method" deprecation warning

joostvanrijn:
When calling `.path` on a record a deprecation warning gets triggered:
`DEPRECATION WARNING: Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s)`

This fixes that. Unfortunately I couldn't run the appraisal tests myself. I couldn't install the `mysql` gem.


xymbol:
Checked on Travis and, the errors reported seem to be network related (and intermittent, most probably). Would be good to request a rebuild. On a 5.2 project this deprecation warning can easily show up all over the place.

kbrock:
kicking

kbrock:
Think the gem doesn't even need this ordering.
Do you all even have the need for this order method going forward?

PR #374, opened 2018-01-07
Fix spelling

fwoelm:
Just a minor change of 'where' to 'were'.

On a side note:

How about renaming the heading **Using acts_as_tree instead of has_ancestry** to **Using has_ancestry instead of acts_as_tree**? Or, IMO even better: **Use has_ancestry instead of acts_as_tree**

The current wording had me confused a bit, I was not sure if it was telling me that I should or should not be using acts_as_tree instead of has_ancestry.

PS: Amazing gem :heart:


fwoelm:
NM -- closed in favor of #368 :)

Issue #372, opened 2017-12-30
Deprecation Warning: The behavior of `attribute_changed?` will be changing.


Flauschbaellchen:
I'm using `self.subtree` in an after_save callback.

While using `Rails 5.1.4` I'm getting this warning:

> DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails.The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead.

My use-case is to track the latest reply-timestamp within a tree of emails:
```ruby
class Letter < ApplicationRecord
  has_ancestry orphan_strategy: :adopt
  after_save :touch_root_letter
  def update_last_reply_at!
    self.update_columns(last_reply_at: self.subtree.where(draft: false).pluck(:created_at).max || self.created_at)
  end
  def touch_root_letter
    self.root.update_last_reply_at!
  end
end
```

Flauschbaellchen:
As soon as I submitted this issue I re-checked the PRs and found #370 which I missed somehow on my first time. Sorry.

kbrock:
Hello @Flauschbaellchen 

Please let me know if I understand this correctly:

If the tree changes, then you want to update the `subtree` based upon the new tree - right?
So it looks like it will work correctly with the newly defined rails behavior.

I think the issue is: There is a deprecation message and we want to double check things and make that deprecation go away.

Sound right to you?

kbrock:
@Flauschbaellchen I put together #387.  I thought I had resolved all these issues in the last release, but a few more was sneaking through.

If you get the chance, could you let me know if your workflow still works with #387 (or maybe it has merged by the time you read this)

Thanks so much for bringing this up to my attention.

causztic:
Hello,

I have this following line of code:
`after_save :update_full_path, if: -> { saved_change_to_name? }`

```
  def update_full_path
    full_path = if root?
                  name
                else
                  (path - [root]).map(&:name).join "/"
                end

    update_column(:full_path, full_path)
    touch
    children.each(&:update_full_path)
  end
```

I am still having the deprecation messages of:

```
[...]

PR #376, opened 2018-02-15
Replace deprecated AR::Dirty for AR >= 5.1

consti:
This PR replaces the deprecated AR::Dirty methods for AR >= 5.1.0.

#370 unfortunately did not fix the issue for me (we have to check if there are `saved_changes?`, otherwise `attribute_before_last_save` will return `nil`).

#372 is related (a duplicate)


consti:
This is still broken. I'll re-open when I'm confident that it is fixed for good.

PR #377, opened 2018-02-16
Replace deprecated AR::Dirty for AR >= 5.1

### LINKED ISSUE #372: Deprecation Warning: The behavior of `attribute_changed?` will be changing.

consti:
This PR replaces deprecated AR::Dirty methods (AR 5.1.0).

This includes more changes than #370 and deals with both states (`saved_changes?` or no saved changes).

This resolves #372

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -78,7 +78,13 @@ module Ancestry
       # New records cannot have children
       raise Ancestry::AncestryException.new('No child ancestry for new record. Save record before performing tree operations.') if new_record?
 
-      if self.send("#{self.ancestry_base_class.ancestry_column}_was").blank? then id.to_s else "#{self.send "#{self.ancestry_base_class.ancestry_column}_was"}/#{id}" end
+      ancestry_value = if ActiveRecord::VERSION::STRING >= '5.1.0'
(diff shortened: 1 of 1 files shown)


kbrock:
darn. looks like the 5.1 code is not returning the expected results (per travis)

consti:
@kbrock yeah, the problem is that ancestry uses some methods on `before_save` and `after_save` / `after_destroy` callbacks - and with Rails 5.1 this requires different methods to access `changed` values.

kbrock:
@consti is what we have on master ok now?
I punted on the whole `changed` array and use `*_was`

PR #379, opened 2018-03-06
Add support for Rails 5.1+ AR Dirty changes

jjuliano:
from `saved_changes?` to `has_changes_to_save?`, `column_was` to `column_in_database` and `changed` to `changed_attribute_names_to_save`. Check version string for each changes for backward compatibility.

### DIFF
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -86,7 +86,7 @@ module Ancestry
       after_destroy :touch_ancestors_callback
 
       if ActiveRecord::VERSION::STRING >= '5.1.0'
-        after_save :touch_ancestors_callback, if: :saved_changes?
+        after_save :touch_ancestors_callback, if: :has_changes_to_save?
       else
         after_save :touch_ancestors_callback, if: :changed?
       end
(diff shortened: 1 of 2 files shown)


lib/ancestry/has_ancestry.rb
```diff
@@ -86,7 +86,7 @@ def has_ancestry options = {}
       after_destroy :touch_ancestors_callback
 
       if ActiveRecord::VERSION::STRING >= '5.1.0'
-        after_save :touch_ancestors_callback, if: :saved_changes?
+        after_save :touch_ancestors_callback, if: :has_changes_to_save?
```
khustochka: `saved_changes?` and `has_changes_to_save?` have different semantics. `saved_changes?` is true if there were any changes when an instance was saved last time. `has_changes_to_save?` (which is an alias to `changed?`) is true if there are _unsaved_ changes. 

With this change touch_ancestors_callback is never called, two more tests fail. Tested with AR 5.2.0.

kbrock:
thank you for diving into this. I'm trying to make some cleanup to make this easier to get fixed

kbrock:
This is great stuff.

- I added appraisals for 5.2
- I changed the `changed` check to just use`_was`
- I got the tests passing (though I'm unsure if they are false positives)
- I used your fix for `dangerous errors`

Could you rebase and then help me understand the various methods that you all suggest instead of using`ancestry_was`?

Thanks for this PR. It is helping me get up to date with 5.2.

ASIDE: I'm looking at replacing `child_ancestry` with `path_ids_was`, `ancestry` with`ancestry_ids` and `ancestry_was` with `ancestry_ids_was` everywhere. Keep that in mind as we are going through this code.

lib/ancestry/has_ancestry.rb
kbrock: @khustochka thanks for the insight. It sounds like `save_changes?` is the one we till want to use

kbrock:
I do have a branch that uses `ids` instead of the `ancestry` strings. but I've separated that.

I did need to revisit this. I've put together #387 instead. Hope that will meet our needs

PR #380, opened 2018-03-07
Correct Internals Documentation

besquared:
Materialized paths are wild-carded from the right hand side according to the code.

https://github.com/stefankroes/ancestry/blob/master/lib/ancestry/materialized_path.rb#L35

Generally speaking b-tree indexes can only do right-hand side LIKE optimizations, I've listed the postgresql documentation below as an example:

https://www.postgresql.org/docs/9.2/static/indexes-types.html


kbrock:
@besquared good catch on the typo.

Just to have fun. On the mac, symbols (i.e.: `/`) are not part of the index. so the indexing is all wonky. Really should be doing a c index here.

PR #383, opened 2018-04-17
use ancestor_ids instead of string munging

kbrock:
Use ancestor_ids more
Move materialized path munging to `MaterializedPath` module.

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -155,12 +155,12 @@ module Ancestry
           # For each node ...
           scope.find_each do |node|
             # ... rebuild ancestry from parents array
-            ancestry, parent = nil, parents[node.id]
+            ancestor_ids, parent = [], parents[node.id]
             until parent.nil?
-              ancestry, parent = if ancestry.nil? then parent else "#{parent}/#{ancestry}" end, parents[parent]
+              ancestor_ids, parent = [parent] + ancestor_ids, parents[parent]
             end
             node.without_ancestry_callbacks do
-              node.update_attribute node.ancestry_column, ancestry
+              node.update_attribute :ancestor_ids, ancestor_ids
             end
           end
         end
@@ -168,13 +168,13 @@ module Ancestry
     end
 
     # Build ancestry from parent id's for migration purposes
-    def build_ancestry_from_parent_ids! parent_id = nil, ancestry = nil
+    def build_ancestry_from_parent_ids! parent_id = nil, ancestor_ids = []
       unscoped_where do |scope|
         scope.where(:parent_id => parent_id).find_each do |node|
           node.without_ancestry_callbacks do
-            node.update_attribute ancestry_column, ancestry
+            node.update_attribute :ancestor_ids, ancestor_ids
           end
-          build_ancestry_from_parent_ids! node.id, if ancestry.nil? then "#{node.id}" else "#{ancestry}/#{node.id}" end
+          build_ancestry_from_parent_ids! node.id, ancestor_ids + [node.id]
         end
       end
     end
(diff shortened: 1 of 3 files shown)


lib/ancestry/instance_methods.rb
```diff
@@ -17,13 +17,7 @@ def update_descendants_with_new_ancestry
           # ... replace old ancestry with new ancestry
           descendant.without_ancestry_callbacks do
             descendant.update_attribute(
-              self.ancestry_base_class.ancestry_column,
-              descendant.read_attribute(descendant.class.ancestry_column).gsub(
-                # child_ancestry_was
-                /^#{self.child_ancestry}/,
-                # future child_ancestry
-                if ancestors? then "#{read_attribute self.class.ancestry_column }/#{id}" else id.to_s end
-              )
```
NickLaMuro: I am still trying to understand what this code did, but I can see why this might have been your rationale for wanted to make this switch a bit more now 😅

lib/ancestry/instance_methods.rb
```diff
@@ -17,13 +17,7 @@ def update_descendants_with_new_ancestry
...
+              :ancestor_ids, path_ids + (descendant.ancestor_ids - path_ids_was)
```
NickLaMuro: Is there a reason this needs to stay multi-line now?  I guess to stay under a character count, so I guess that is cool.

Possibly consider doing:

```ruby
updated_ids = path_ids + (descendant.ancestor_ids - path_ids_was)
descendant.update_attribute(:ancestor_ids, updated_ids)
```

But that is just how I tend to organize code.

lib/ancestry/class_methods.rb
```diff
@@ -155,26 +155,26 @@ def restore_ancestry_integrity!
           # For each node ...
           scope.find_each do |node|
             # ... rebuild ancestry from parents array
-            ancestry, parent = nil, parents[node.id]
+            ancestor_ids, parent = [], parents[node.id]
             until parent.nil?
-              ancestry, parent = if ancestry.nil? then parent else "#{parent}/#{ancestry}" end, parents[parent]
+              ancestor_ids, parent = [parent] + ancestor_ids, parents[parent]
```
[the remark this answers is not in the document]
kbrock: I tried making that change, and got an error:
```
NoMethodError: undefined method `id' for 1:Integer
```

Which is good for 2 reasons.

1. that means this code is actually tested
2. that means that parent contains an integer (the id) and not an object.

Should I assume a followup reponse that `parent` should be renamed to `parent_id`?

lib/ancestry/instance_methods.rb
kbrock: thanks for this comment - think I introduced `ancestry_ids_was` - when I should be going towards the whole `ancestor_ids_before_last_save` route

lib/ancestry/instance_methods.rb
kbrock: I'm doing this change right now - thanks

lib/ancestry/class_methods.rb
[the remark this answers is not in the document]
NickLaMuro: > Should I assume a followup response that `parent` should be renamed to `parent_id`?

I guess if you are pushing for a new version change, I guess that might not be a bad idea since that is the generally the convention, but I wouldn't say it is a requirement now knowing that this method breaks the convention.

lib/ancestry/instance_methods.rb
NickLaMuro: Well, I don't know that comment was worth thanking me for, I just understand the rational a bit better now that I have had the time to understand it a bit better.

PR #384, opened 2018-04-17
Fix 'dangerous errors' when using in Rails 5.2

kbrock:
- deprecation errors taken from #379 thanks @jjuliano 
- just using `ancestry_was` for now to make tests pass
- add rails to build matrix

Just trying to make this pass for now. Not sure if it is actually working with rails 5.2


lib/ancestry/has_ancestry.rb
```diff
@@ -46,7 +46,7 @@ def has_ancestry options = {}
       scope :siblings_of, lambda { |object| where(sibling_conditions(object)) }
       scope :ordered_by_ancestry, Proc.new { |order|
         if %w(mysql mysql2 sqlite sqlite3 postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::MAJOR >= 5
-          reorder("coalesce(#{connection.quote_table_name(table_name)}.#{connection.quote_column_name(ancestry_column)}, '')", order)
+          reorder(Arel.sql("coalesce(#{connection.quote_table_name(table_name)}.#{connection.quote_column_name(ancestry_column)}, '')"), order)
         else
           reorder("(CASE WHEN #{connection.quote_table_name(table_name)}.#{connection.quote_column_name(ancestry_column)} IS NULL THEN 0 ELSE 1 END), #{connection.quote_table_name(table_name)}.#{connection.quote_column_name(ancestry_column)}", order)
```
mu-is-too-short: You might want to include an `Arel.sql` call in this branch too.

mu-is-too-short: It'll go into the `CASE` version for any driver that isn't MySQL, SQLite, or PostgreSQL regardless of the ActiveRecord major version.

That said, that's an odd looking test to me, the AR version shouldn't matter and I'm pretty sure every database these days has a `coalesce` function. But maybe I'm wrong and this is just bike-shedding anyway

PR #386, opened 2018-04-26
trim object allocations while parsing ancestry col

kbrock:
tested for `#ancestor_ids` (since parsing is main part of that method

value      |      3.0.2|      after| ∆
-----------|-----------|-----------|---
ips        |128,935.154|176,705.314|37%
ips σ      |     20.107|     13.874| :+1:
allocations|         24|         14|42%

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -206,5 +206,24 @@ module Ancestry
         yield self.ancestry_base_class.unscope(:where)
       end
     end
+
+    ANCESTRY_UNCAST_TYPES = [:string, :uuid, :text].freeze
+    if ActiveSupport::VERSION::STRING < "4.0"
+      def primary_key_is_an_integer?
+        if defined?(@primary_key_is_an_integer)
+          @primary_key_is_an_integer
+        else
+          @primary_key_is_an_integer = !ANCESTRY_UNCAST_TYPES.include?(columns_hash[primary_key.to_s].type)
+        end
+      end
+    else
+      def primary_key_is_an_integer?
+        if defined?(@primary_key_is_an_integer)
+          @primary_key_is_an_integer
+        else
+          @primary_key_is_an_integer = !ANCESTRY_UNCAST_TYPES.include?(type_for_attribute(primary_key))
+        end
+      end
+    end
   end
 end
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -93,10 +93,6 @@ module Ancestry
       changed.include?(self.ancestry_base_class.ancestry_column.to_s)
     end
 
-    def parse_ancestry_column obj
-      obj.to_s.split('/').map { |id| cast_primary_key(id) }
-    end
-
     def ancestor_ids
       parse_ancestry_column(read_attribute(self.ancestry_base_class.ancestry_column))
     end
@@ -288,16 +284,10 @@ module Ancestry
 
   private
 
-    def cast_primary_key(key)
-      if [:string, :uuid, :text].include? primary_key_type
-        key
-      else
-        key.to_i
-      end
-    end
-
-    def primary_key_type
-      @primary_key_type ||= column_for_attribute(self.class.primary_key).type
+    def parse_ancestry_column obj
+      obj_ids = obj.to_s.split('/')
+      obj_ids.map!(&:to_i) if self.class.primary_key_is_an_integer?
+      obj_ids
     end
 
     def unscoped_descendants


lib/ancestry/instance_methods.rb
```diff
@@ -288,16 +284,21 @@ def ancestry_callbacks_disabled?
...
+    end
+
+    def skip_cast?
+      return @skip_cast if defined?(@skip_cast)
+      @skip_cast ||= [:string, :uuid, :text].include?(primary_key_type)
     end
 
     def primary_key_type
-      @primary_key_type ||= column_for_attribute(self.class.primary_key).type
+      column_for_attribute(self.class.primary_key).type
```
NickLaMuro: Confused why you got rid of the memoization here.  Doesn't seem like it would help, but only hurt.

lib/ancestry/instance_methods.rb
```diff
@@ -288,16 +284,21 @@ def ancestry_callbacks_disabled?
...
-        key
-      else
-        key.to_i
-      end
+    def parse_ancestry_column obj
+      cast_primary_keys(obj.to_s.split('/'))
+    end
+
+    def cast_primary_keys(keys)
+      skip_cast? ? keys : keys.map(&:to_i)
```
NickLaMuro: Can this be a `.map!` to avoid allocating a new array?

lib/ancestry/instance_methods.rb
```diff
@@ -288,16 +284,21 @@ def ancestry_callbacks_disabled?
...
+    end
+
+    def skip_cast?
+      return @skip_cast if defined?(@skip_cast)
+      @skip_cast ||= [:string, :uuid, :text].include?(primary_key_type)
```
NickLaMuro: You should make `[:string, :uuid, :text]` a constant to avoid allocating that every time this method is called (if you were doing this over multiple records, this would not be memoized every time).

lib/ancestry/instance_methods.rb
NickLaMuro: If you decide to do the above, maybe change the name of the method to `cast_primary_keys!` to denote that it mutates that array.

lib/ancestry/instance_methods.rb
kbrock: this is only used in 1 place (to determine if we should skip keys) - so decided to only keep 1 in memory. although, it is probably pretty minor

lib/ancestry/class_methods.rb
```diff
@@ -206,5 +206,14 @@ def unscoped_where
         yield self.ancestry_base_class.unscope(:where)
       end
     end
+
+    ANCESTRY_UNCAST_TYPES = [:string, :uuid, :text].freeze
+    def primary_key_is_an_integer?
```
NickLaMuro: I really like what you ended up with here.  👍

lib/ancestry/instance_methods.rb
```diff
@@ -288,16 +284,10 @@ def ancestry_callbacks_disabled?
...
-      if [:string, :uuid, :text].include? primary_key_type
-        key
-      else
-        key.to_i
-      end
-    end
-
-    def primary_key_type
-      @primary_key_type ||= column_for_attribute(self.class.primary_key).type
+    def parse_ancestry_column obj
```
NickLaMuro: Nice!  You were able to remove a method and is still pretty readable.  Both good things!  👍

lib/ancestry/class_methods.rb
NickLaMuro: Though, looking at the test failures, you might have to do something like this if we are still supporting Rails 3:

```ruby
if Rails.version < "4.0"
  def primary_key_is_an_integer?
    if defined?(@primary_key_is_an_integer)
      @primary_key_is_an_integer
    else
      @primary_key_is_an_integer = !ANCESTRY_UNCAST_TYPES.include?(columns_hash[primary_key.to_s].type)
    end
  end
else
  # The method definition you have here
end
```

The `columns_hash[primary_key.to_s].type` is the class method equivalent of `column_for_attribute`.

Actually, you should just do that for all versions, since it seems to exist in Rails 5 as well.

kbrock: I started to change it but just using the pre 4.0 version:

```ruby
    ANCESTRY_UNCAST_TYPES = [:string, :uuid, :text].freeze
    if ActiveSupport::VERSION::STRING < "4.0"
      def primary_key_is_an_integer?
        if defined?(@primary_key_is_an_integer)
          @primary_key_is_an_integer
        else
          @primary_key_is_an_integer = !ANCESTRY_UNCAST_TYPES.include?(column_for_attribute(primary_key).type)
        end
      end
    else
      def primary_key_is_an_integer?
        if defined?(@primary_key_is_an_integer)
          @primary_key_is_an_integer
        else
          @primary_key_is_an_integer = !ANCESTRY_UNCAST_TYPES.include?(type_for_attribute(primary_key))
        end
      end
    end
```

kbrock:
@NickLaMuro Thanks for pointers on failing tests. I thought the answer could use less internals but alas, your solution was the way to go.

PR #391, opened 2018-05-12
small fix with nested scopes using object id

fursich:


fursich:
Oops, I pushed a wrong branch. Sorry.

Issue #390, opened 2018-05-12
chained scopes not working properly


fursich:
Just encountered a minor errors related with chained scopes when working on some other implementations (#388)

### Here's a test case to reproduce:

```ruby
# test/concerns/scopes_test.rb

  def test_chained_scopes
    AncestryTestDatabase.with_model :depth => 2, :width => 1 do |model, roots|
      root = roots.first.first

        assert_equal model.children_of(root.id).children_of(root.id).to_a, model.children_of(root.id).to_a
    end
  end
```

Intuitively we would expect to get `children of the root`, if we scope `children of the root` out of themselves, however, this raises errors as below.

```ruby
Loaded Ancestry test suite environment:
  Ruby: 2.5.0
  ActiveRecord: 4.2.10
  Database: SQLite

Run options: --seed 37168

# Running:

...................................................E..................

Finished in 3.373931s, 20.7473 runs/s, 669.2490 assertions/s.

  1) Error:
ScopesTest#test_chained_scopes:
ActiveRecord::RecordNotFound: Couldn't find AncestryTestDatabase::TestChainedScopesTestNode with 'id'=1 [WHERE "test_nodes"."ancestry" = '1']
```

This occurs because (if I understand correctly) `root.id` in the second scope turns `nil` when evaluated within the first scope's context. (see [here](https://github.com/stefankroes/ancestry/blob/60782fdd744100126de5e443a53632152ab89f96/lib/ancestry/class_methods.rb#L5) )

### PR

As I personally need it to be fixed soon, I'm very happy to submit a quick PR for it with a few test cases added... Hope that helps!

fursich:
closing this as the fix proposal has been merged

PR #392, opened 2018-05-13
allow chanied scopes to handle object.id properly

fursich:
This is a small fix for #390.

currently, chained scopes as below raises errors:

```ruby
model.where.not(id: root).children_of(root.id)
```

because the latter `children_of(root.id)` (or any other scopes in ancestry) calles `find(root.id)`, which is being executed under the earlier scopes and become equivalent to `find(nil)`.

Tests are added with a few comments on ActiveRecord version dependency of this behavior.


fursich:
Just realized `unscoped_find` is available. (commits combined as one)
Hope that makes sense!

PR #408, opened 2018-09-05
fix reverse ordering

### LINKED ISSUE #312: IrreversibleOrderError
In Rails 5.0.0.1

Sth like this `child.ancestors.last` produces this:

> Finding a last element by loading the relation when SQL ORDER can not be reversed is deprecated. Rails 5.1 will raise ActiveRecord::IrreversibleOrderError in this case. Please call `to_a.last` if you still want to load the relation.

Of course i don't want to fetch all relation to get its last record.

kbrock:
creating `Ascending` is the trick here

fixes #312

@simonc if you are still using ancestry, could you let me know if this fixes your issue? thnx


simonc:
Hi @knrock. Thanks for the fix! From what I see it seems to indeed fix our issue. We played with our data to avoid the issue until now but we won't have to anymore 😊

PR #410, opened 2018-09-24
Scope for indirects

KevinColemanInc:
This is a dupe of #358, but I rebased it to catch up with master and added more testing. I do not think tilo is maintaining his pr any more.


KevinColemanInc:
This is a dupe of #358, but I rebased it to catch up with master

kbrock:
@KevinColemanInc

Looks like you and I duplicated work with #420. Sorry I didn't see yours first

Thank you for your help and let me know if I did something different from what you were thinking.
--K

PR #412, opened 2018-10-26
Fix primary_key integer check

adam101:
The latest release (3.0.3) has broken our code, because of https://github.com/stefankroes/ancestry/pull/386/.

The issue is the following:

```
pry:1> ActiveSupport::VERSION::STRING
=> "4.2.8"
pry:1> self.class.type_for_attribute(self.class.primary_key)
=> #<ActiveRecord::Type::String:0x007ffac181f628 @limit=24, @precision=nil, @scale=nil>
pry:1> self.class.primary_key_is_an_integer?
=> true
pry:1> self.id
=> "5bd1da2c4624b83bd51e40af"
pry:1> self.id.to_i
=> 5
```

Because of the ActiveSupport version, it uses the `type_for_attribute` function which returns with an `ActiveRecord::Type::String` instance which is not included in the `ANCESTRY_UNCAST_TYPES` array. What included is the `type` of this instance.

```
[78] pry:1> self.class.type_for_attribute(self.class.primary_key).type
=> :string
```

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -208,22 +208,16 @@ module Ancestry
     end
 
     ANCESTRY_UNCAST_TYPES = [:string, :uuid, :text].freeze
-    if ActiveSupport::VERSION::STRING < "4.0"
-      def primary_key_is_an_integer?
-        if defined?(@primary_key_is_an_integer)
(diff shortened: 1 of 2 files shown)


kbrock:
@adam101 thanks so much for this fix. I merged #413 instead (gave you full attribution)

PR #419, opened 2018-11-02
Don't drop parentless nodes

### LINKED ISSUE #362: .arrange method filters out some records.

### LINKED ISSUE #411: sort_by_ancestry should not require the full tree
If you pass `sort_by_ancestry` an array of nodes that doesn't include the whole tree (or perhaps all the nodes on a branch), it doesn't return all the nodes you passed it. For example, this test fails:

```ruby
def test_sort_by_ancestry_with_incomplete_tree
  AncestryTestDatabase.with_model do |model|
    # - n1
    #   - n2
    #     - n3
    #   - n4
    n1 = model.create!
    n2 = model.create!(parent: n1)
    n3 = model.create!(parent: n2)
    n4 = model.create!(parent: n1)
    assert_equal [n3, n4].map(&:id), model.sort_by_ancestry([n3, n4]).map(&:id)
  end
end
```

I suspect this something to do with 8a4fb0083db246ca8e3462db148bba202b5c8b4c, in which @Fryguy replaced the existing implementation of `arrange_nodes` with one that is more efficient. @Fryguy, I have to admit I don't entirely understand how your implementation works, but do you think it could be modified so that the above test would pass?

kbrock:
Extracted from #415 

Fixes #411
Fixes #362

`arrange_nodes` no longer drops child nodes. thanks @trafium

@Fryguy Think this is the simplified version of 415 that you wanted

We also now have other tests showing what other work needs to be done

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -38,22 +38,16 @@ module Ancestry
 
     # Arrange array of nodes into a nested hash of the form
     # {node => children}, where children = {} if the node has no children
+    # If a node's parent is not included, the node will be included as if it is a top level node
     def arrange_nodes(nodes)
-      arranged = ActiveSupport::OrderedHash.new
-      nodes_ids = nodes.map(&:id)
-      min_depth = nodes.map(&:depth).min
-      index = Hash.new { |h, k| h[k] = ActiveSupport::OrderedHash.new }
+      node_ids = Set.new(nodes.map(&:id))
+      index = Hash.new { |h, k| h[k] = {} }
 
-      nodes.each do |node|
+      nodes.each_with_object({}) do |node, arranged|
         children = index[node.id]
         index[node.parent_id][node] = children
-
-        if node.depth == min_depth || !nodes_ids.include?(node.parent_id)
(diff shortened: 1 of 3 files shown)


Fryguy:
:+1:

PR #431, opened 2019-03-13
Minor update to which versions of Rails are supported

Zeragamba:
Our team is going though the process of upgrading Rails and ran into the issue that type_for_attribute was undefined. After some investigation, we found that type_for_attribute was added in Rails 4.2 and was not defined previously.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ records.
 
 NOTE:
 
-- Ancestry 3.x supports Rails 5.0 and earlier.
+- Ancestry 2.x supports Rails 4.1, and earlier
+- Ancestry 3.x supports Rails 5.0, and 4.2
 - Ancestry 4.0 only supports rails 5.0 and higher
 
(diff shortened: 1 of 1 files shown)


kbrock:
Thank you @SpyMaster356 and darn. Was trying so hard not to drop support for 4.1 until later on.
Then again, I probably dropped test support assuming that not much was changing in the older versions of the code. :(

Do you need ancestry 3.x changes for rails 4.1?
Are you planning on sticking with rails 4.1?
Or are you transitioning through it on your journey to rails 5.x and doing a large transition?

SpyMaster356:
Heh, ya sorry to be the messenger of bad news.

For now we've down graded to 2.2.2 while we continue our journey to Rails 5 and beyond.

kbrock:
Thanks for sharing.
Best of luck.
Let us know if we can help you in your journey to a recent rails

PR #432, opened 2019-03-17
Test against Ruby 2.5 and 2.6

yuki24:


kbrock:
hmm... this doesn't seem to be running through travis

@stefankroes is this still configured for travis?
Is it time to move this to it's own domain? maybe `ancestry-ruby` ? (ancestry is owned by ancestry.com)

kbrock:
@yuki24 I agree that dropping 2.4 and testing against 2.5 makes sense. But I'm unsure what testing against 2.6 buys us.

We are testing against 3 (soon to be 4 rails versions) - It is tricky to work across these versions and since we push the envelope here, then that makes sense here.  I've even dropped a few versions of active record sooner than I would have like because it was getting too difficult to install older versions of the old mysql driver.

We're also testing across mysql, sqlite, and postgres. These also are different and cause subtle issues.

My concern stems around adding 3 ruby versions, resulting in 36 tests. That feels a little excessive.

I'm leaning towards just testing against ruby 2.5, but would like your feedback.
Is there a particular concern around these versions of ruby?

Thank you,
Keenan

kbrock:
thank you

PR #434, opened 2019-03-28
WIP - Fix rails 5.1.6 deprecation on changed.

jrafanie:
Fixes rails 5.1.6 deprecation:

DEPRECATION WARNING: The behavior of `changed_attributes` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_changes.transform_values(&:first)` instead. (called from ancestry_changed? at .../2.5.0/gems/ancestry-3.0.5/lib/ancestry/instance_methods.rb:104)

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -139,7 +139,11 @@ module Ancestry
     alias :has_parent? :ancestors?
 
     def ancestry_changed?
-      changed.include?(self.ancestry_base_class.ancestry_column.to_s)
+      if ActiveRecord::VERSION::STRING >= '5.1.0'
+        will_save_change_to_attribute?(self.ancestry_base_class.ancestry_column.to_s)
+      else
+        changed.include?(self.ancestry_base_class.ancestry_column.to_s)
+      end
     end
 
     def ancestor_ids


lib/ancestry/instance_methods.rb
```diff
@@ -139,7 +139,11 @@ def ancestors?
     alias :has_parent? :ancestors?
 
     def ancestry_changed?
-      changed.include?(self.ancestry_base_class.ancestry_column.to_s)
+      if ActiveRecord::VERSION::STRING >= '5.1.0'
+        will_save_change_to_attribute?(self.ancestry_base_class.ancestry_column.to_s)
```
kbrock: @jrafanie thanks

looks like this in 2017, `will_save_change_to_attribute` changes again.
In 5.1, does `saved_change_to_attribute?` also meet your needs?
Or should we put a double if there to support the newer versions

jrafanie: I'll have to take a look. Thanks

jrafanie: Ok, I took a look further.  `saved_changed_to_attribute?` is for after save callbacks to indicate we just saved changes to a specific attribute.  Before save callbacks much ust the will_save...

see https://www.ombulabs.com/blog/rails/upgrades/active-record-5-1-api-changes.html

jrafanie: oh wait, it looks like this method is called from both before and after save callbacks.  I'll have to fix this.

jrafanie: I need to find a cleaner way of implementing this.  It's fixed in two different places for "before save" using "will save changes" and "after save" "saved changes to".

jrafanie:
Opened #441 as a replacement

PR #442, opened 2019-05-27
Update codebase to remove Rails 6 deprecation warnings

d-staehler:
Hey! We're planning on upgrading to Rails 6.0.0.rc1 soon. I traced a number of deprecation warnings back to `ancestry`

These include using `update_attributes` over `update` as well as using class level querying methods if the receiver scope has leaked (see https://github.com/rails/rails/pull/35280)

One thing I am not 100% sure about is using `unscoped` over `unscope(:where)`. Feedback would be appreciated on this.

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -202,7 +202,7 @@ module Ancestry
           yield self.ancestry_base_class
         end
       else
-        yield self.ancestry_base_class.unscope(:where)
+        yield self.ancestry_base_class.unscoped
       end
     end
 


kbrock:
Thank you @d-staehler 

having trouble getting travis running in this repo. will ping @stefankroes again to hopefully resolve it.

possibly `except(:where)` will work instead of `unscoped(:where)` - I'll look at that

d-staehler:
Hey @kbrock,

Were you able to get Travis running for ancestry? I am also happy to test out the difference between `except(:where)` and `unscoped(:where)` if you haven't had the chance to yet. 

Cheers

lib/ancestry/class_methods.rb
```diff
@@ -198,7 +198,7 @@ def unscoped_where
           yield self.ancestry_base_class
         end
       else
-        yield self.ancestry_base_class.unscope(:where)
+        yield self.ancestry_base_class.unscoped
```
kbrock: I'm not sold that this line is causing the deprecation warnings

kbrock:
updated gemfiles
cherry picked the `s/update_attributes/update/`

haven't done the `unscoped` yet, I also tried `excepts`, but neither seem to resolve the deprecation messages.

This is how I run locally:

```
bundle
bundle exec appraisal install
# wwtd --local
DB=sqlite3 BUNDLE_GEMFILE=gemfiles/gemfile_60.gemfile bundle exec rake
```

--Kb

d-staehler:
Hey @kbrock, 

Thanks for that! I just pointed to the latest changes on `master` and I am still getting the following when I run RSpec tests:

```
DEPRECATION WARNING: Class level methods will no longer inherit scoping from `create!` in Rails 6.1. To continue using the scoped relation, pass it into the block directly. To instead access the full set of models, as Rails 6.1 will, use `LibraryCategory.unscoped`. (called from unscoped_where at /vendor/bundle/ruby/2.5.0/gems/ancestry-3.0.7/lib/ancestry/class_methods.rb:201)
```

It definitely has issues with the `create` statements as per the above. I do not see the above deprecation warning with `unscoped`

kbrock:
@d-staehler I think the fix is to convert the `unscoped_where` from a block to a simple scope

not totally sure how this will work.

Also, I know we don't test against active record 3.2 - but a little hesitant to change things and totally drop it. I may need to bite the bullet soon and just update the minimum rails version

hidde-jan:
> Also, I know we don't test against active record 3.2 - but a little hesitant to change things and totally drop it. I may need to bite the bullet soon and just update the minimum rails version.

3.2 was EOL'ed in 2015. That's four years ago. And if you don't even test, then you can't be sure that you even actually support it.

kbrock:
Just made some larger changes. got rid of  a `_was` too.

am circling back to see if I can change the unscoped code easily enough.

(if you could rebase on master when you get the chance, that would be great)

richard-viney:
Thanks, after rebasing there is now only one line left in this PR's diff. It is to do with the use of `unscoped` mentioned previously. I'm not 100% certain if this change is correct, would need to investigate more closely to see what the desired behaviour actually is here.

Also, there are two deprecation warnings running the tests for Rails 6:

```
DEPRECATION WARNING: Class level methods will no longer inherit scoping from `create!` in Rails 6.1. To continue using the scoped relation, pass it into the block directly. To instead access the full set of models, as Rails 6.1 will, use `AncestryTestDatabase::TestNodeCreationInAfterCommitTestNode.unscoped`. (called from children_of at /Users/RichardLSG/code/ancestry/lib/ancestry/materialized_path.rb:37)
DEPRECATION WARNING: Class level methods will no longer inherit scoping from `create!` in Rails 6.1. To continue using the scoped relation, pass it into the block directly. To instead access the full set of models, as Rails 6.1 will, use `AncestryTestDatabase::TestNodeCreationInAfterCommitTestNode.unscoped`. (called from children_of at /Users/RichardLSG/code/ancestry/lib/ancestry/materialized_path.rb:37)
```

Thanks!

kbrock:
Looks like rails wants us to use `scoping`. Which is unfortunate since ancestry has always advertised `children.create`.

```diff
- model.children.create
+ model.children.scoping { model.create }
```

An alternative would be to convert `children` from a scope (technically a static method) to an actual relation. But I can't figure out how to do that (it would be optimal if someone can suggest a way to do that)

kbrock:
kicking

mmoll:
What's the status here?

kbrock:
@mmoll this does not pass tests. if you know how to solve, please share

mmoll:
@kbrock could you re-kick the travis tests for current master? At least for me, these also seem to fail on pg.

mmoll:
@kbrock GH-476 for fixing Travis.

mmoll:
needs a rebase now :)

kbrock:
@mmoll yes, these all pass. but they still produce the deprecation warning as advertised.

rails has trouble with our `children.create` and would prefer we use `model.children.scoping { model.create }`.  I'm not sure the best way around this

brendon:
We had a similar issue here. To be honest I don't think we have solved it yet but I'll link you just in case there's something there that will help: https://github.com/brendon/acts_as_list/pull/363

kbrock:
thank you for all the help here. From this PR we got in a lot of good stuff.

Thanks @cmr119 for getting us over the finish line. it is just a little different from the suggested change here. 

thanks to both of your for your help

Issue #446, opened 2019-06-25
Getting error after update from 3.0.4 to 3.0.7


AllanKlaus:
I was updating my app and started to get error when update from 3.0.4 to  3.0.7.

My problem is because i'm working with a legacy database and the ancestry was created as a integer. I can't  update the database because other application use it.

the error is

```
NoMethodError:
        undefined method `split' for 13654:Fixnum
      # /home/klaus/.rvm/gems/ruby-2.3.1/gems/ancestry-3.0.7/lib/ancestry/instance_methods.rb:381:in `parse_ancestry_column'
      # /home/klaus/.rvm/gems/ruby-2.3.1/gems/ancestry-3.0.7/lib/ancestry/instance_methods.rb:153:in `ancestor_ids'
```

I saw the in the version `v3.0.5` was changed the method `parse_ancestry_column` and removed the `to_s` on obj. I think it's a breaking change, and cant be done in a patch update. 

I created a pull request to return the `to_s` and avoid this breaking change.

kbrock:
merged

PR #447, opened 2019-06-25
Fix error when ancestry field is integer

AllanKlaus:
- Problem
When the field field ancestry is an integer on database the method
`parse_ancestry_column` raise error because of split is not a method
of integer

- Solution
Return the method `to_s` on the method `parse_ancestry_column` to change
integer to string before do the split

- Error Log

```
NoMethodError:
        undefined method `split' for 13654:Fixnum
      # /home/klaus/.rvm/gems/ruby-2.3.1/gems/ancestry-3.0.7/lib/ancestry/instance_methods.rb:381:in `parse_ancestry_column'
      # /home/klaus/.rvm/gems/ruby-2.3.1/gems/ancestry-3.0.7/lib/ancestry/instance_methods.rb:153:in `ancestor_ids'
```

- Issue: #446

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -377,7 +377,7 @@ module Ancestry
 
     def parse_ancestry_column obj
       return [] unless obj
-      obj_ids = obj.split(ANCESTRY_DELIMITER)
+      obj_ids = obj.to_s.split(ANCESTRY_DELIMITER)
       self.class.primary_key_is_an_integer? ? obj_ids.map!(&:to_i) : obj_ids
     end
 


kbrock:
Hello @AllanKlaus 
I'm a little confused how the `ancestry` column can be an integer.

It needs to hold a path "id/id/id" - how can that be stored in an integer field?

AllanKlaus:
@kbrock it's a legacy project with more then 20 years. First it's  created using PHP, in 2013 was migrated to Ruby. The gem was added to project 5 years ago. I'm in the project about 3 years. 

I think in the begining who created the relations in the project starter using manual relations, but after they changed to use the ancestry and didn't change the database. In the way the gem was doing the relations until version 3.0.4 using `to_s` didn't broke the project. 

I can't not change the database because others legacy applications consult it.

kbrock:
@AllanKlaus I appreciate wanting to keep the database the same.

I'm not asking you to change your database. I'm actually asking you if you even want ancestry.
From the very first version of ancestry, the `ancestry` column contains `ancestors.join("/")`

So I'm surprised that this ever worked.
I guess it could work if you only ever had 2 layers of nodes.

I have been looking into using an `integer[]` or an `ltree`, but support for those are not even close.

Just an integer field can't contain multiple levels. So it seems to go against the whole premise of ancestry and materialized path pattern.

kbrock:
I'll merge, but your use case is not a supported one.

PR #449, opened 2019-08-15
Allow string ids

dennym:
There were tries in the past to implement this feature but way to complicated through different issues and PRs. No idea why you would limit the regex to just integer ids in the beginning. But all the other solutions bloat the code to achieve a simple functionality.

#254 #262 #265 

This PR would make it possible to drop the option `primary_key_format` for the majority of cases that don't use integer ids. It still could be required if someone uses special characters for ids (no idea if this is even possible). If there is no reason against the changing the regex I suggest roll this simple solution.

Addtionally `primary_key_format` is documented in the readme but not reflected in the code.

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -2,7 +2,7 @@ module Ancestry
   module ClassMethods
     # Fetch tree node if necessary
     def to_node object
-      if object.is_a?(self.ancestry_base_class) then object else unscoped_where{|scope| scope.find object} end
+      if object.is_a?(self.ancestry_base_class) then object else unscoped_where{|scope| scope.find object.id} end
     end
 
     # Scope on relative depth options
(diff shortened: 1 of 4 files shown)


lib/ancestry/class_methods.rb
```diff
@@ -2,7 +2,7 @@ module Ancestry
   module ClassMethods
     # Fetch tree node if necessary
     def to_node object
-      if object.is_a?(self.ancestry_base_class) then object else unscoped_where{|scope| scope.find object} end
+      if object.is_a?(self.ancestry_base_class) then object else unscoped_where{|scope| scope.find object.id} end
```
kbrock: great change - thanks

kbrock:
Thanks - cherry-picked "use id in find instead of object"

(did tweak it so tests passed)

kbrock:
@dennym let me know if this doesn't meet your needs

PR #451, opened 2019-09-06
fix :primary_key_format

### LINKED ISSUE #221: Primary id column restriction? Please confirm

kbrock:
Thanks for all the feed back all
And sorry for taking so many years to get this in

This should make it easier for developers to support the various schemes

I'm tempted to completely remove this field, but this solution looked simple enough.

Was tempted to auto derive the field but it connects to the database at class definition time, so thew that away (I'll add the code as reference in a comment for archival purposes)

In the documentation, the delimiter is not specified. But since a few schemes want to specify the delimiter (ltree), I wanted that to be an option. (probably should be doing that when I actually need it... sorry)

fix #221 and #241

See also #132 #262 #265

### DIFF
--- a/lib/ancestry.rb
+++ b/lib/ancestry.rb
@@ -6,5 +6,4 @@ require_relative 'ancestry/has_ancestry'
 require_relative 'ancestry/materialized_path'
 
 module Ancestry
-  ANCESTRY_PATTERN = /\A[0-9]+(\/[0-9]+)*\Z/
 end
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -4,7 +4,7 @@ module Ancestry
       # Check options
       raise Ancestry::AncestryException.new("Options for has_ancestry must be in a hash.") unless options.is_a? Hash
       options.each do |key, value|
-        unless [:ancestry_column, :orphan_strategy, :cache_depth, :depth_cache_column, :touch, :counter_cache].include? key
+        unless [:ancestry_column, :orphan_strategy, :cache_depth, :depth_cache_column, :touch, :counter_cache, :primary_key_format].include? key
           raise Ancestry::AncestryException.new("Unknown option for has_ancestry: #{key.inspect} => #{value.inspect}.")
         end
       end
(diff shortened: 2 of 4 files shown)


kbrock:
While I liked how it worked without developer interaction, it connected to the database - so it was out.

Do feel all of this is stuff is too smart anyway.

```ruby
    def derive_ancestry_pattern(primary_key_format)
      unless primary_key_format
        primary_key_format = if [:string, :text].include?(type_for_attribute(primary_key).type)
                               '[a-zA-Z]+'
                             elsif type_for_attribute(primary_key).type == :uuid
                               '[-0-9a-fA-F]+'
                             else # integer
                                '[0-9]+'
                             end
      end
      if primary_key_format.to_s.include?('\A')
        primary_key_format
      else
        delimiter = '/'
        /\A#{primary_key_format}(#{delimiter}#{primary_key_format})*\Z/
      end
```

tgrushka:
Removed in 2012, put back in master in 2018, still not released in 2020... ?

tgrushka:
Still, integrity check does not work:
```ruby
has_ancestry primary_key_format: /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/

...

Model.build_ancestry_from_parent_ids!
Model.check_ancestry_integrity!

#=> Ancestry::AncestryIntegrityException (Invalid format for ancestry column of node 95ab7e5f-11c7-41c6-8c9a-87d6ad28789f: 0d5bfece-8e2f-4caf-912f-12e5e3b8ab4d.)
```

monsterooo:
> Removed in 2012, put back in master in 2018, still not released in 2020... ?
> 
> EDIT: Not ideal, but this is how I got this functionality in my Gemfile:
> 
> ```ruby
> gem 'ancestry', github: 'stefankroes/ancestry', ref: 'beea5b7'
> ```

Why was it not released

PR #462, opened 2019-10-01
Add graph examples in documentation

dtamai:
## Context

In my last job we used ancestry in a core model of our app, so it was common having questions about tree navigation. I usually draw some example graphs to explain and in my experience it is easy to understand trees visually.

This PR adds some visual documentation for the navigation methods. Below is the rendered table.

image

I chose a thick border to indicate the node from which the navigation method is called and a second color to indicate the nodes returned by the method, I hope this is enough to understand the examples, and I accept suggestions for other schemas.

## The images

For the images I used [mermaid](https://github.com/knsv/mermaid), they are the _mmd_ files in this PR. And to generate the images I used [mermaid-cli](https://github.com/mermaidjs/mermaid.cli) because it creates PNG from the mmd files in a single command.

The bash command to generate all images is:

```bash 
# in img folder
for mmd in *.mmd; do mmdc -i "$mmd" -o "${mmd%.mmd}.png" -b transparent; done
```


dtamai:
> @dtamai I see these images again and again in the documentation and they are super clear and helpful.
> 
> I just revisited today wondering if I wanted to create more images for some concepts.
> It was easily find this PR and the command to generate the images is right there.
> 
> Thank you again for this PR.

Really glad to have been helpful! :smile:

PR #463, opened 2019-10-02
Add tests for indirects in TreeNavigationTest

dtamai:
## Context

When I was checking how the navigation methods behave for #462 I was using the tests in TreeNavigationTest file, then I found out there were no tests for the `indirects` method.

For completeness I added tests for `subtree_ids` and `indirects` and `indirect_ids`.


kbrock:
thanks

Issue #282, opened 2016-06-16
Bulk update parent?


solars:
Hi,

Maybe I didn't find it but is there a way to bulk update the parent of a bunch of objects? (update_all)
Theoretically it's just a matter of mapping the ancestry string and forwarding to update_all, is that right?

Thanks,
Christoph

stefankroes:
I think the parent has a method `child_ancestry`, you can update the ancestry of multiple records to that value using update_all.

kbrock:
@solars were you able to resolve this?

solars:
Not yet, still having a workaround in place

fwoelm:
If you are using PostgreSQL, you might even be able to bulk update all descendants:

``` ruby
tree_to_move = Tree.find(100)
# ID: 100
# Ancestry Path: 1/2/3

new_parent_tree = Tree.find(1000)
# ID: 1000
# Ancestry Path: 4

# Move all of the tree node's children:
# Replace in ancestry_path (the first occurrence of):
# tree_to_move's child ancestry path (1/2/3/100)
# with
# the child ancestry path of the new parent + the ID of the tree we are moving (4/1000/100)
# so that...
# 1/2/3/100 becomes 4/1000/100
# 1/2/3/100/101 becomes 4/1000/100/101
# 1/2/3/100/102 becomes 4/1000/100/102
# ...
tree_to_move.descendants.update_all("ancestry = REGEXP_REPLACE(ancestry, '#{tree_to_move.child_ancestry}', '#{new_parent_tree.child_ancestry}/#{tree_to_move.id}')")

# Last but not least we have to move the tree node itself (we only updated descendants so far).
# Use update_column to avoid triggering the has_ancestry callback that manually updates each
# descendant record. 
tree_to_move.update_column('ancestry',  new_parent_tree.child_ancestry)
```
I have only done some preliminary testing but moving a tree with 8420 descendants (20 children with each 20 children with each 20 children) went from 16s (individual update statements) to .5s (update_all). A 32x improvement looks pretty promising :smile: It also reduces the risk of race conditions.

But: More work is needed on this.

PS: If you are not using IDs, make sure to look into string escaping for the ancestry path in the #update_all statement to make sure you are not opening yourself up to injections or unexpected behavior.

kbrock:
I have been focused on changing the path from a ruby string concatenation  to something more sql friendly like an array. This is the case that I would like to target.

[...]

PR #464, opened 2019-10-16
A Way to optimize for Postgres

suonlight:
# Summary

Currently, there is no way to make optimization for a specific database.
This PR is to give an approach to help a developer use database techniques to do optimization.
For example, the origin of `update_descendants_with_new_ancestry` has a lot of N+1 queries to update all descendants. By using `regexp_replace` on Postgres, we can use one query to remove N+1.

Related Issue: https://github.com/stefankroes/ancestry/issues/282

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ rvm:
   - 2.6
 
 env:
+  - DB=pg OPTIMIZER=true
   - DB=pg
   - DB=mysql2
   - DB=sqlite3
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -24,6 +24,10 @@ module Ancestry
       # Include instance methods
       include Ancestry::InstanceMethods
 
+      if Ancestry.optimizer
+        include Ancestry.optimizer
+      end
+
       # Include dynamic class methods
       extend Ancestry::ClassMethods
 
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -24,9 +24,7 @@ module Ancestry
       # Include instance methods
       include Ancestry::InstanceMethods
 
-      if Ancestry.optimizer
-        include Ancestry.optimizer
-      end
+      include Ancestry.optimizer if Ancestry.optimizer
 
       # Include dynamic class methods
       extend Ancestry::ClassMethods
--- a/lib/ancestry/optimizer.rb
+++ b/lib/ancestry/optimizer.rb
@@ -1,0 +1,9 @@
+module Ancestry
+  class << self
+    attr_reader :optimizer
+
+    def optimizer= klass
+      @optimizer ||= klass
+    end
+  end
+end
--- a/lib/ancestry/optimizers/pg.rb
+++ b/lib/ancestry/optimizers/pg.rb
@@ -1,0 +1,26 @@
+module Ancestry
+  module Optimizers
+    module Pg
+      # Update descendants with new ancestry (before save)
+      def update_descendants_with_new_ancestry
+        # If enabled and node is existing and ancestry was updated and the new ancestry is sane ...
+        if !ancestry_callbacks_disabled? && !new_record? && ancestry_changed? && sane_ancestry?
+          column = self.ancestry_base_class.ancestry_column.to_s
+          old_ancestry = self.child_ancestry
+          new_ancestry = ancestors? ? "#{read_attribute self.class.ancestry_column }/#{id}" : id.to_s
+          update_clause = [
+            "#{column} = regexp_replace(#{column}, '^#{old_ancestry}', '#{new_ancestry}')"
+          ]
+          if self.ancestry_base_class.respond_to?(:depth_cache_column) && self.respond_to?(self.ancestry_base_class.depth_cache_column)
+            column = self.ancestry_base_class.depth_cache_column.to_s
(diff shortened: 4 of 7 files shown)


lib/ancestry/has_ancestry.rb
```diff
@@ -24,6 +24,10 @@ def has_ancestry options = {}
       # Include instance methods
       include Ancestry::InstanceMethods
 
+      if Ancestry.optimizer
+        include Ancestry.optimizer
```
thukim: minor: `include Ancestry.optimizer if Ancestry.optimizer`

lib/ancestry/optimizer.rb
```diff
@@ -0,0 +1,9 @@
+module Ancestry
+  class << self
+    attr_reader :optimizer
+
+    def optimizer= klass
+      @optimizer ||= klass
```
thukim: 👍

lib/ancestry/optimizers/pg.rb
```diff
@@ -0,0 +1,26 @@
+module Ancestry
+  module Optimizers
+    module Pg
+      # Update descendants with new ancestry (before save)
+      def update_descendants_with_new_ancestry
+        # If enabled and node is existing and ancestry was updated and the new ancestry is sane ...
+        if !ancestry_callbacks_disabled? && !new_record? && ancestry_changed? && sane_ancestry?
+          column = self.ancestry_base_class.ancestry_column.to_s
+          old_ancestry = self.child_ancestry
```
thukim: `self` is redundant

test/environment.rb
```diff
@@ -118,6 +119,23 @@ def self.create_test_nodes model, depth, width, parent = nil
       end
     else; []; end
   end
+
+  private
+
+  def self.use_optimizer
```
kbrock: yea, think this can be dropped / moved up to `has_ancestry`

lib/ancestry/optimizers/pg.rb
```diff
@@ -0,0 +1,27 @@
+module Ancestry
+  module Optimizers
+    module Pg
+      # Update descendants with new ancestry (before save)
+      def update_descendants_with_new_ancestry
```
kbrock: Also, could you look at the new way we do `update_descendants_with_new_ancestry` and see that it still looks like the best approach now.

I've moved over to using arrays instead of all the string munging we were doing before.

Think using the update clause to update all could be a great optimization.

Do wonder about objects already in memory.

lib/ancestry/optimizers/pg.rb
```diff
@@ -0,0 +1,27 @@
+module Ancestry
+  module Optimizers
+    module Pg
```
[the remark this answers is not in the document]
suonlight: Right now, Rails 6 supports multiple databases and ActiveRecord has a way to establish new db connection. If there is a use case for switching between mysql and pg, then this approach will breaks. I'm thinking about changing it into class, and then we use a factory method to get the right optimizer for pg/mysql/... at runtime. It means `InstanceMethods#update_descendants_with_new_ancestry` change to:

```
def update_descendants_with_new_ancestry
   if optimizer
     optimizer.update_descendants_with_new_ancestry
  else
     # origin methods
  end
end
```

But I feel like it is still not good,  it is harder to maintain. Do you have any suggestions?

lib/ancestry/optimizers/pg.rb
suonlight: I can not use arrays because the generated SQL for `update_all` does not care about each descendant's ancestor_ids `new_ancestor_ids = path_ids + (descendant.ancestor_ids - path_ids_in_database)`. It just uses `regexp_replace` to update all descendants. So I have to use a string concatenation for `update_all`.

> Do wonder about objects already in memory.

descendants need to be reloaded to get new values. I'll update and test again.

lib/ancestry/optimizers/pg.rb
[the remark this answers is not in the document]
kbrock: I think the database type will probably be hard coded at design time, and definitely at deploy time.
It will not change at runtime.

But I do see pointing to serverless and a relational at the same time.
Find it doubtful they would point to mysql and postgres at the same time.

lib/ancestry/has_ancestry.rb
```diff
@@ -24,6 +24,8 @@ def has_ancestry options = {}
       # Include instance methods
       include Ancestry::InstanceMethods
 
+      include Ancestry.optimizer if Ancestry.optimizer
```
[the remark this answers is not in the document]
kbrock: if nothing else, move below MaterializedPath

test/environment.rb
```diff
@@ -118,6 +119,23 @@ def self.create_test_nodes model, depth, width, parent = nil
...
+    use_optimizer = ENV['OPTIMIZER'] == 'true'
+    return unless use_optimizer
+
+    if db_type == 'pg'
+      require File.expand_path('../../lib/ancestry/optimizers/pg', __FILE__)
```
kbrock: I think this can be required regardless of the database type.

We just won't include it unless we want the optimization

kbrock:
thank you for these suggestions and the code.

I tweaked it a little but it is basically your code and I think it will help many people make this faster.

.travis.yml
```diff
@@ -7,6 +7,7 @@ rvm:
   - 2.6
 
 env:
+  - DB=pg OPTIMIZER=true
```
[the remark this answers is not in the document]
kbrock: to get this add to an initializer:

```ruby
Ancestry.default_update_strategy = :sql
```

as seen in [[environment.rb]](https://github.com/stefankroes/ancestry/blob/077d4b/test/environment.rb#L52)

PR #467, opened 2019-11-27
Sort improvement

dukaev:
Hi!
Thank you for gem.
I added small performance improvement for sort
[`sort vs sort_by`](https://github.com/JuanitoFatas/fast-ruby#enumerablesort-vs-enumerablesort_by-code)

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -72,7 +72,7 @@ module Ancestry
       arranged = nodes if nodes.is_a?(Hash)
 
       unless arranged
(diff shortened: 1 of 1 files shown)


kbrock:
hello @dukaev 

Do these tests pass for you?
in ruby 2.5.5 all versions of rails (and all dbs) fail for me

I appreciate why you would want `sort` over `sort_by`, but the block in this method (part of existing api before I inherited the project) makes it difficult to change.

Were you improving a particular issue, or were you just applying the well known performance optimization?

kbrock:
kicking build

PR #476, opened 2020-04-23
Fix DB detection for Travis build

mmoll:


mmoll:
I'm unsure what implication this has on local builds, but theoretically this should have been broken since 95dcbe93c0082020c95463e11e041d7807610338.

kbrock:
thanks

PR #480, opened 2020-06-18
cache ancestor_ids

d-m-u:
This is intended as a start to the PR for moving the remainder of the ancestor_ids caching that miq does currently. 

There will need to be a few changes if it needs to be working on the 3.0.7 version. 
The `parent_id` method in particular will need to change on since `cast_primary_key` is recent.

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -7,14 +7,19 @@ module Ancestry
 
     # Update descendants with new ancestry (before save)
     def update_descendants_with_new_ancestry
-      # If enabled and node is existing and ancestry was updated and the new ancestry is sane ...
-      if !ancestry_callbacks_disabled? && !new_record? && ancestry_changed? && sane_ancestry?
-        # ... for each descendant ...
-        unscoped_descendants.each do |descendant|
-          # ... replace old ancestry with new ancestry
-          descendant.without_ancestry_callbacks do
-            new_ancestor_ids = path_ids + (descendant.ancestor_ids - path_ids_in_database)
-            descendant.update_attribute(:ancestor_ids, new_ancestor_ids)
+      # If callbacks enabled
+      if !ancestry_callbacks_disabled?
+        clear_memoized_instance_variables
+        # if node is existing and ancestry was updated and the new ancestry is sane ...
+        if !new_record? && ancestry_changed? && sane_ancestry?
+          # ... for each descendant ...
+          unscoped_descendants.each do |descendant|
+            # ... replace old ancestry with new ancestry
+            descendant.without_ancestry_callbacks do
+              new_ancestor_ids = path_ids + (descendant.ancestor_ids - path_ids_in_database)
+              descendant.update_attribute(:ancestor_ids, new_ancestor_ids)
+            end
+            descendant.clear_memoized_instance_variables
           end
         end
       end
@@ -158,9 +163,35 @@ module Ancestry
     end
 
     def parent_id
-      ancestor_ids.last if ancestors?
+      return @_parent_id if defined?(@_parent_id)
+      @_parent_id = if @_ancestor_ids
+                      @_ancestor_ids.empty? ? nil : @_ancestor_ids.last
+                    else
+                      col = read_attribute(ancestry_base_class.ancestry_column)
+                      # Specifically not using `.blank?` here because it is
+                      # slower than doing the below.
+                      if col.nil? || col.empty? # rubocop:disable Rails/Blank
+                        nil
+                      else
+                        rindex = col.rindex(Ancestry::MaterializedPath::ANCESTRY_DELIMITER)
+                        cast_primary_key(rindex ? col[rindex + 1, col.length] : col)
+                      end
+                    end
     end
     alias :parent_id? :ancestors?
+
+    def depth
+      @_depth ||= if @_ancestor_ids
+                    @_ancestor_ids.size
+                  else
+                    col = read_attribute(ancestry_base_class.ancestry_column)
+                    col ? col.count(Ancestry::MaterializedPath::ANCESTRY_DELIMITER) + 1 : 0
+                  end
+    end
+
+    def cast_primary_key(key)
+      self.class.primary_key_is_an_integer? ? key.to_i : key
+    end
 
     def parent
       unscoped_find(parent_id) if ancestors?
@@ -288,7 +319,15 @@ module Ancestry
       defined?(@disable_ancestry_callbacks) && @disable_ancestry_callbacks
     end
 
-  private
+    def clear_memoized_instance_variables
+      @_ancestor_ids = nil
+      @_depth        = nil
+
+      # can't assign to `nil` since `nil` could be a valid result
+      remove_instance_variable(:@_parent_id) if defined?(@_parent_id)
+    end
+
+    private
     def unscoped_descendants
       unscoped_where do |scope|
         scope.where self.ancestry_base_class.descendant_conditions(self)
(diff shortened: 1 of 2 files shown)


lib/ancestry/materialized_path.rb
```diff
@@ -97,6 +97,16 @@ def sane_ancestry?
         (ancestry_value.nil? || !ancestor_ids.include?(self.id)) && valid?
       end
 
+      def parent= parent
```
kbrock, line 100: think this will take care of itself if you change `ancestor_ids=`

kbrock: but not sure if you want to implement `ancestry=`
that seems a good place to mess up our cache

lib/ancestry/instance_methods.rb
```diff
@@ -151,14 +151,30 @@ def ancestor_of?(node)
     # assuming that parent hasn't changed
     def parent= parent
       self.ancestor_ids = parent ? parent.path_ids : []
+      clear_memoized_instance_variables
```
kbrock, line 154: think you want to do this work over in ancestry_ids=

d-m-u: I'm not sure I believe you.

lib/ancestry/materialized_path.rb
```diff
@@ -106,10 +106,15 @@ def ancestors?
       def ancestor_ids=(value)
         col = self.ancestry_base_class.ancestry_column
         value.present? ? write_attribute(col, value.join(ANCESTRY_DELIMITER)) : write_attribute(col, nil)
+        @_ancestor_ids = nil
+        @_depth        = nil
+
+        # can't assign to `nil` since `nil` could be a valid result
+        remove_instance_variable(:@_parent_id) if defined?(@_parent_id)
```
jgigault, line 113: I guess that "ancestor_ids=" should return assigned value to allow "var = record. ancestor_ids = 'smth'" ?

lib/ancestry/instance_methods.rb
```diff
@@ -1,5 +1,7 @@
 module Ancestry
   module InstanceMethods
+    ANCESTRY_DELIMITER = '/'.freeze unless defined?(ANCESTRY_DELIMITER)
```
kbrock, line 3: is this necessary? I think this is defined elsewhere

lib/ancestry/instance_methods.rb
```diff
@@ -151,17 +155,45 @@ def ancestor_of?(node)
     # assuming that parent hasn't changed
     def parent= parent
       self.ancestor_ids = parent ? parent.path_ids : []
+      clear_memoized_instance_variables
     end
 
     def parent_id= new_parent_id
       self.parent = new_parent_id.present? ? unscoped_find(new_parent_id) : nil
+      clear_memoized_instance_variables
```
kbrock, line 163: move to `ancestor_ids=`

lib/ancestry/instance_methods.rb
```diff
@@ -133,10 +141,6 @@ def path depth_options = {}
       self.ancestry_base_class.scope_depth(depth_options, depth).ordered_by_ancestry.inpath_of(self)
     end
 
-    def depth
```
kbrock, line 136: please keep here - it will get overridden in subclass

lib/ancestry/instance_methods.rb
```diff
@@ -151,17 +155,45 @@ def ancestor_of?(node)
...
     end
 
     def parent_id= new_parent_id
       self.parent = new_parent_id.present? ? unscoped_find(new_parent_id) : nil
+      clear_memoized_instance_variables
     end
 
     def parent_id
-      ancestor_ids.last if ancestors?
+      return @_parent_id if defined?(@_parent_id)
```
kbrock, line 167: wondering if we need this optimization.

Asked @NickLaMuro and he is evaluating if we loose too much performance by dropping the `@_parent_id` optimization.

lib/ancestry/instance_methods.rb
```diff
@@ -151,17 +155,45 @@ def ancestor_of?(node)
...
+          nil
+        else
+          rindex = col.rindex(ANCESTRY_DELIMITER)
+          cast_primary_key(rindex ? col[rindex + 1, col.length] : col)
+        end
+      end
     end
     alias :parent_id? :ancestors?
 
+    def depth
```
kbrock, line 184: ditto from above

lib/ancestry/instance_methods.rb
```diff
@@ -151,17 +155,45 @@ def ancestor_of?(node)
...
+      @_depth ||= if @_ancestor_ids
+        @_ancestor_ids.size
+      else
+        col = read_attribute(ancestry_base_class.ancestry_column)
+        col ? col.count(ANCESTRY_DELIMITER) + 1 : 0
+      end
+    end
+
+    def cast_primary_key(key)
```
kbrock, line 193: I could have sworn that this was no longer necessary.

aah - it is still used in parent_id - if we drop that optimization, then we drop this method

lib/ancestry/materialized_path.rb
kbrock: maybe move line 109-113 between 106/107- so the write attribute is returning the right value.

thanks @jgigault

lib/ancestry/instance_methods.rb
```diff
@@ -18,6 +18,12 @@ def update_descendants_with_new_ancestry
           end
         end
       end
+      unless ancestry_callbacks_disabled?
```
NickLaMuro, line 21: So, in my previous implementation:

https://github.com/ManageIQ/manageiq/blob/6b9b4724c6fb63017a44d4917239cf6806698dea/lib/patches/ancestry_patch.rb#L2-L10

The extra `unless` was needed since I needed to call this after `super`.

Now that we are working within a single implementation (and not a monkey patch), we can probably refactor this entire method to share the `if` that is already there:

```ruby
def update_descendants_with_new_ancestry
  # If callbacks enabled
  if !ancestry_callbacks_disabled?
    clear_memoized_instance_variables
    # if node is existing and ancestry was updated and the new ancestry is sane ...
    if !new_record? && ancestry_changed? && sane_ancestry?
      # ... for each descendant ...
      unscoped_descendants.each do |descendant|
        # ... replace old ancestry with new ancestry
          new_ancestor_ids = path_ids + (descendant.ancestor_ids - path_ids_in_database)
          descendant.update_attribute(:ancestor_ids, new_ancestor_ids)
        end
        descendant.clear_memoized_instance_variables
      end
    end
  end
end
```

lib/ancestry/instance_methods.rb
```diff
@@ -151,17 +157,45 @@ def ancestor_of?(node)
...
     def parent_id= new_parent_id
       self.parent = new_parent_id.present? ? unscoped_find(new_parent_id) : nil
+      clear_memoized_instance_variables
     end
 
     def parent_id
-      ancestor_ids.last if ancestors?
+      return @_parent_id if defined?(@_parent_id)
+      @_parent_id = if @_ancestor_ids
+        @_ancestor_ids.empty? ? nil : @_ancestor_ids.last
```
NickLaMuro, line 171: Oh, you just really wanted to trigger my OCD (not real) with this indenting here, huh?

lib/ancestry/instance_methods.rb
```diff
@@ -151,17 +157,45 @@ def ancestor_of?(node)
...
+          rindex = col.rindex(Ancestry::MaterializedPath::ANCESTRY_DELIMITER)
+          cast_primary_key(rindex ? col[rindex + 1, col.length] : col)
+        end
+      end
     end
     alias :parent_id? :ancestors?
 
+    def depth
+      @_depth ||= if @_ancestor_ids
+        @_ancestor_ids.size
```
NickLaMuro, line 188: You monster...

d-m-u:
hey @kbrock what else do you need from me here?

kbrock:
@d-m-u sorry to do this but do you think we should close this?

PR #481, opened 2020-06-20
[POC] rails attribute for ancestor_ids

kbrock:
Rails has a concept of serializing attributes. They even allow you to define a custom serializer.

So the concept of `ancestry` and `ancestor_ids` are merged, and the delimiter basically goes away.

So from there, I merged them over to the `ancestor_ids` concept and saw if it was possible to rename fields based upon ancestor_ids. If this is possible, it may be possible to have multiple ancestors.

Dropping 4.2 to simplify this support.

### DIFF
--- a/lib/ancestry.rb
+++ b/lib/ancestry.rb
@@ -71,4 +71,76 @@ module Ancestry
   def self.default_primary_key_format=(value)
     @@default_primary_key_format = value
   end
+
+  # TODO: move out of here
+  # used for materialized path
+  class MaterializedPathString < ActiveRecord::Type::Value
+    def initialize(casting: :to_i, delimiter: '/')
+      @casting = casting&.to_proc
+      @delimiter = delimiter
+    end
+
+    def type
+      :materialized_path_string
+    end
+
+    # convert to database type
+    def serialize(value)
+      if value.kind_of?(Array)
+        value.map(&:to_s).join(@delimiter).presence
+      elsif value.kind_of?(Integer)
+        value.to_s
+      elsif value.nil? || value.kind_of?(String)
+        value
+      else
+        byebug
+        puts "curious type: #{value.class}"
+      end
+    end
+
+    def cast(value)
+      cast_value(value) #unless value.nil? (want to get rid of this - fix default value)
+    end
+
+    # called by cast (form or setter) or deserialize (database)
+    def cast_value(value)
+      if value.kind_of?(Array)
+        super
+      elsif value.nil?
+        # would prefer to use default here
+        # but with default, it kept thinking the field had changed when it hadn't
+        super([])
+      else
+        #TODO: test ancestry=1
+        super(value.to_s.split(@delimiter).map(&@casting))
+      end
+    end
+  end
+end
+ActiveRecord::Type.register(:materialized_path_string, Ancestry::MaterializedPathString)
+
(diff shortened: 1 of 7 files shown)


lib/ancestry.rb
```diff
@@ -6,4 +6,74 @@
...
+    # convert to database type
+    def serialize(value)
+      if value.kind_of?(Array)
+        value.map(&:to_s).join(@delimiter).presence
+      elsif value.kind_of?(Integer)
+        value.to_s
+      elsif value.nil? || value.kind_of?(String)
+        value
+      else
+        byebug
```
d-m-u, line 29: i think we want to handle this case a little differently

kbrock: lol

kbrock:
this turns out to not be trivial to rebase and bring forward. frustrating
have tried a few times and no dice.
especially frustrating since this really doesn't touch anything that changed that much.

the issue I have here is rails 5.1(?) and earlier don't handle the serialization that well - refusing to cache it and not properly detecting when the value has changed. Well, to be more precise, not detecting when the value has not changed. It also does not cache the casting so quite a few more objects are created.

I would prefer a way to make the attribute look more like a rails database parameter. The integration points are close but not quite there in rails proper. Requires just a little too much monkey patching.

i do have a modified version of activerecord-virtual_attributes that is more native. but again, it is close but still requires too many monkey patches.

kbrock:
so it looks like there are a few things to do for this

Determine whether we want to expose `ancestry` as a string for compatibility, or we want to expose as an array, and use `ancestry_before_type_cast` to represent the serialized value. Or if we want to use `ancestor_ids` as the core concept and throw `ancestry` in there for legacy purposes.

- get away from directly manipulating the serialized version of the ancestor_ids (aka `ancestry` field) - it is an implementation detail on how the tree of ancestors was serialized to disk.
- serialize and deserialize the `ancestor_ids` using rails.
- introduce a method to define these methods instead of including a module. This is the way that rails does most of its magic with attribute accessors.

kbrock:
stalled: ancestry column becomes an array. so the concept of ancestor_ids and ancestry merge. Which I guess makes sense because the slash delimiter was just a database serialization artifact.

problem: lots of people expect the ancestry column to have a slash delimiter. So having it converted to an array will mess with them. Not sure how to be backwards compliant in this concept

lib/ancestry.rb
```diff
@@ -34,4 +34,75 @@ def self.default_update_strategy
...
+    end
+
+    def type
+      :materialized_path_string
+    end
+
+    # convert to database type
+    def serialize(value)
+      if value.kind_of?(Array)
+        value.map(&:to_s).join(@delimiter).presence
```
kbrock, line 52: not sure if we want an empty ancestry to have `[]` or `nil`

technically null means that the value is undetermined. But in our case, we want to say "none" which is "" and not nil.

kbrock:
our main code base has move to using `ancestor_ids` in most places.
This helps us since having rails use a serializer or a a method don't matter as long as we are using `int[]` to access the columns.

I have had trouble bringing this up to date. Have tried a few times.

kbrock:
this is imposible to rebase.
I have already pulled many of these concepts out of here.

Keeping around but will need to completely rewrite this when I get the chance

PR #487, opened 2020-07-09
Better humanise model name for error message

mkllnk:
Imagine a model called `ProductType`. The old message was:

> Producttype cannot be a descendant of itself.

The new message is:

> Product type cannot be a descendant of itself.

There is probably an even better way of making the whole message translatable but I didn't look into that.


kbrock:
thanks

Issue #486, opened 2020-07-08
Simplify ordering SQL for MySQL


brendon:
Hi there :) I was wondering if you'd consider simplifying the ordering SQL for MySQL instances:

https://github.com/stefankroes/ancestry/blob/master/lib/ancestry/materialized_path.rb#L78

MySQL orders NULL FIRST by default (for ASC. NULL LAST for DESC) so we don't need to worry about an extra SQL for MySQL:
```ruby
    if %w(mysql mysql2).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::MAJOR >= 5
      reorder(arel_table[ancestry_column])
    elsif %w(sqlite sqlite3 postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::MAJOR >= 5
      reorder(
        Arel::Nodes::Ascending.new(Arel::Nodes::NamedFunction.new('COALESCE', [arel_table[ancestry_column], Arel.sql("''")])),
          order
        )
      else
        reorder(Arel.sql("(CASE WHEN #{connection.quote_table_name(table_name)}.#{connection.quote_column_name(ancestry_column)} IS NULL THEN 0 ELSE 1 END), #{connection.quote_table_name(table_name)}.#{connection.quote_column_name(ancestry_column)}"), order)
      end
```

It feels nicer not to make MySQL do extra work when it's not necessary. What do you think?

brendon:
Sorry I see these have been moved out of scopes anyway in `master`. I am getting warnings about overwriting your scopes as I have simplified these two scopes in my model, but using your new version I'd imagine I could just overwrite those methods locally for the same effect.

Do you have a timeline on when you'll be doing a new release? :)

kbrock:
@brendon I have a lot of documentation for the changes in ancestry. I'll see what is left to do and see about releasing soon

Do you know if the forward and reverse sorting works for mysql?
Postgres possibly has sorting with nulls in some cases, so maybe we could get both of those situations with a single call

also, the next release moves a few of the methods around, so some monkey patches may have issues.

brendon:
MySQL sorts NULLS FIRST in ASC and NULLS LAST in DESC so a predictable pattern. I’ve heard Postgres is not so predictable and also differs per operating system. Probably best to leave the fix on there for Postgres or perhaps probe the DB for the nulls first setting and throw a warning if it will cause problems?

brendon:
[...]

PR #490, opened 2020-07-14
simplify mysql ordering

d-m-u:
was asked to open by KB, it's stolen directly from https://github.com/stefankroes/ancestry/issues/486
thanks @brendon 

per that issue, MySQL's default is NULL FIRST; we don't care about activesupport 4.2 any more so the version check can be removed, and the else clause is also no longer necessary (see https://github.com/stefankroes/ancestry/pull/490#discussion_r456938446)

### DIFF
--- a/lib/ancestry/materialized_path.rb
+++ b/lib/ancestry/materialized_path.rb
@@ -75,13 +75,15 @@ module Ancestry
     end
 
     def ordered_by_ancestry(order = nil)
-      if %w(mysql mysql2 sqlite sqlite3 postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::MAJOR >= 5
+      if %w(mysql mysql2 sqlite sqlite3).include?(connection.adapter_name.downcase)
+        reorder(arel_table[ancestry_column], order)
+      elsif %w(postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::STRING >= "6.1"
+        reorder(Arel::Nodes.new(arel_table[ancestry_column]).nulls_first)
+      else
         reorder(
           Arel::Nodes::Ascending.new(Arel::Nodes::NamedFunction.new('COALESCE', [arel_table[ancestry_column], Arel.sql("''")])),
           order
         )
-      else
-        reorder(Arel.sql("(CASE WHEN #{connection.quote_table_name(table_name)}.#{connection.quote_column_name(ancestry_column)} IS NULL THEN 0 ELSE 1 END), #{connection.quote_table_name(table_name)}.#{connection.quote_column_name(ancestry_column)}"), order)
       end
     end
 


lib/ancestry/materialized_path.rb
```diff
@@ -75,7 +75,9 @@ def siblings_of(object)
     end
 
     def ordered_by_ancestry(order = nil)
-      if %w(mysql mysql2 sqlite sqlite3 postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::MAJOR >= 5
+      if %w(mysql mysql2).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::MAJOR >= 5
```
brendon, line 78: I think that you can include sqlite in this treatment of the column. You probably also don't need the ActiveRecord version check as we're not doing anything fancy with ordering here.

lib/ancestry/materialized_path.rb
```diff
@@ -75,7 +75,9 @@ def siblings_of(object)
...
+        reorder(arel_table[ancestry_column])
+      elsif %w(sqlite sqlite3 postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::MAJOR >= 5
         reorder(
           Arel::Nodes::Ascending.new(Arel::Nodes::NamedFunction.new('COALESCE', [arel_table[ancestry_column], Arel.sql("''")])),
```
brendon, line 82: Instead of `COALESCE` you can order as per the mysql version but add `NULLS FIRST` at the end of the ordering statement. You may need to use arel to build the ordering statement so that it contains the table name to make this work with more complex queries that AR can generate sometimes.

lib/ancestry/materialized_path.rb
kbrock: I agree.

I think that version check is to avoid using an arel named function in the `reorder()` method call

Since the statement is quite simple, we can avoid that.

lib/ancestry/materialized_path.rb
kbrock: For all versions of postgres (again no version check), I believe you can do the same thing:

```suggestion
arel_table[ancestry_column].nulls_first
```
alternatives:
```
arel_table[ancestry_column].asc.nulls_first
Arel::Nodes::Ascending.new(arel_table[ancestry_column]).nulls_first
```

kbrock:
per conversation with @d-m-u - sqlite3 is the same as mysql, nulls come before values.

postgresql has nulls after values. which is tricky when we reverse ordering, but not concerned about that right now especially since we want the values to come back in a certain order

lib/ancestry/materialized_path.rb
```diff
@@ -75,10 +75,11 @@ def siblings_of(object)
     end
 
     def ordered_by_ancestry(order = nil)
-      if %w(mysql mysql2 sqlite sqlite3 postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::MAJOR >= 5
+      if %w(mysql mysql2 sqlite sqlite3).include?(connection.adapter_name.downcase)
+        reorder(arel_table[ancestry_column])
```
kbrock, line 79:
```suggestion
        reorder(arel_table[ancestry_column], order)
```

lib/ancestry/materialized_path.rb
```diff
@@ -75,10 +75,11 @@ def siblings_of(object)
...
+      elsif %w(sqlite sqlite3 postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::MAJOR >= 5
```
kbrock, line 80:
```suggestion
      elsif %w(postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::STRING >= "6.1"
```

This is in rails master. I'm assuming the next version is 6.1

lib/ancestry/materialized_path.rb
brendon: I like `arel_table[ancestry_column].asc.nulls_first` but `asc` is probably not necessary given it's the default.

I didn't realise one could pass `nulls_first` with arel like that!

brendon:
Looking good. Are you able to explain the AR > 6.1 requirement? And is the AR >= 5 requirement still necessary also?

brendon:
I see what you mean. `reverse_order` may not be usable in the case of Postgres with the NULLS FIRST option unless that is also reversed by the method? I've had a quick look at the `.nulls_first` method has only been added to Arel recently. In that case, perhaps it'd be better to hard code it in the ordering string instead?

https://github.com/stefankroes/ancestry/issues/312

The error is being thrown because of the `ActiveRecord::VERSION::STRING >= "6.1"` constraint which makes pg based code fall back to the `IF THEN ELSE` statement.

kbrock:
Do we care about the ordering of nodes?

I think the original purpose is to have `arrange` work. We've since changed it so it should work without any sorting.

The sorting also helps out for pagination. You need a consistent ordering for pagination - though, I'm not sold that we do that. Also, since the hidden nodes on other pages can alter the ordering, a partial result currently can display an incorrect order. Other frameworks like act_as_list (yours?) and closure_trees do better.

`NULLS FIRST` needs to change to `NULLS LAST` when reversing the order.
But I'm not sure what reversing the order does for us. Since we're displaying trees, and need the front nodes to build the tree, what does reversing the order even mean? I'm not sure what we would want from the results.

I'm pretty sure I put in that test because a user's app was blowing up when introducing reversing.

We could undo the postgres/arel changes here and leave the sqlite/mysql changes.
That would move us forward so we don't have to get all existential.

lib/ancestry/materialized_path.rb
```diff
@@ -75,11 +75,10 @@ def siblings_of(object)
...
-        reorder(
-          Arel::Nodes::Ascending.new(Arel::Nodes::NamedFunction.new('COALESCE', [arel_table[ancestry_column], Arel.sql("''")])),
-          order
-        )
+      if %w(mysql mysql2 sqlite sqlite3).include?(connection.adapter_name.downcase)
+        reorder(arel_table[ancestry_column], order)
+      elsif %w(postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::STRING >= "6.1"
+        reorder(Arel::Nodes::Ascending.new(arel_table[ancestry_column]).nulls_first)
       else
         reorder(Arel.sql("(CASE WHEN #{connection.quote_table_name(table_name)}.#{connection.quote_column_name(ancestry_column)} IS NULL THEN 0 ELSE 1 END), #{connection.quote_table_name(table_name)}.#{connection.quote_column_name(ancestry_column)}"), order)
```
kbrock, line 83: ok, thinking this one through.

Before this change, the 3 vendors worked.
Now after this change, which was targeting mysql, postgres is broken.

So just adding line 78 and 79 to the top (and 80-81) would fix this.

We have the coalesce and case statements because sybase, oracle, and microsoft sql server were holdouts for the coalesce statement. They had `ISNULL` or `NVL` or other to do the same thing and it took a while for the ANSI stander to get into the various platforms. IBM DB2 was probably one of the first converts.

This is many many years later, so since we don't care about active support 4.2, I'm thinking we can drop our `else` block (83) and use the previous code for postgres (deleted 79-82)

```suggestion
        reorder(
          Arel::Nodes::Ascending.new(Arel::Nodes::NamedFunction.new('COALESCE', [arel_table[ancestry_column], Arel.sql("''")])),
          order
        )
```

The final diff should probably show this addition as unchanged rather than an addition.

brendon:
It is a conundrum RE `NULLS FIRST` and `LAST`. Ideally Rails or Arel would pick up on its presence and switch it around when `reverse` is called. We can't do much from the outside sadly unless we override `reverse` on classes that use `ancestry`.

In regards to ordering at all, I use it like this if I know that my query won't produce orphans:

`Things.ordered_by_ancestry_and(:position)` or something like that. I don't have to `arrange` them in this instance. `position` comes from `acts_as_list`.

d-m-u:
hey @kbrock what else do you need from me here?

PR #492, opened 2020-07-30
cleanup the readme 

d-m-u:
we don't need two tables and some of this information isn't necessary

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,10 +3,10 @@
 # Ancestry
 
 Ancestry is a gem that allows the records of a Ruby on Rails
-ActiveRecord model to be organised as a tree structure (or hierarchy). It uses
-a single database column, employing the materialised path pattern. It exposes all the standard tree structure
-relations (ancestors, parent, root, children, siblings, descendants) and allows all
-of them to be fetched in a single SQL query. Additional features are STI
+ActiveRecord model to be organised as a tree structure (or hierarchy). It employs
+the materialised path pattern and exposes all the standard tree structure
+relations (ancestors, parent, root, children, siblings, descendants), allowing all
+of them to be fetched in a single SQL query. Additional features include STI
 support, scopes, depth caching, depth constraints, easy migration from older
 gems, integrity checking, integrity restoration, arrangement of
 (sub)trees into hashes, and various strategies for dealing with orphaned
@@ -14,14 +14,13 @@ records.
 
 NOTE:
 
-- Ancestry 2.x supports Rails 4.1, and earlier
-- Ancestry 3.x supports Rails 5.0, and 4.2
+- Ancestry 2.x supports Rails 4.1 and earlier
+- Ancestry 3.x supports Rails 5.0 and 4.2
 - Ancestry 4.0 only supports rails 5.0 and higher
 
 # Installation
 
-To apply Ancestry to any `ActiveRecord` model, follow these simple steps:
-
+Follow these simple steps to apply Ancestry to any ActiveRecord model:
 
 ## Install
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -94,100 +94,73 @@ node.children.create :name => 'Stinky'
...
+      </td>
+      <td>
+         <p align="center">root</p>
+         <img src="img/root.png" alt="root"/>
+         <div align="center"><sub>returns self for a root node</sub></div>
+      </td>
+      <td>
+         <p align="center">ancestors</p>
+         <img src="img/ancestors.png" alt="ancestors"/>
+         <div align="center"><sub>starting with root, ending with parent</sub></sub></div>
```
kbrock, line 116: unlike above, this is a description.
see if this makes it any better

```suggestion
         <div align="center"><sub>root..parent</sub></sub></div>
```

README.md
```diff
@@ -94,100 +94,73 @@ node.children.create :name => 'Stinky'
...
-    <td>
-      <p align="center">subtree</p>
-      <img src="img/subtree.png" alt="subtree"/>
-    </td>
-  </tr>
+   <tr>
+      <td>
+         <p align="center">parent</p>
+         <img src="img/parent.png" alt="parent"/>
+         <div align="center"><sub>returns nil for a root node</sub></div>
```
kbrock, line 106: note: this is a footnote

(that was a fact not a request for change)

README.md
```diff
@@ -94,100 +94,73 @@ node.children.create :name => 'Stinky'
...
+         <p align="center">ancestors</p>
+         <img src="img/ancestors.png" alt="ancestors"/>
+         <div align="center"><sub>starting with root, ending with parent</sub></sub></div>
+      </td>
+   </tr>
+   <tr>
+      <td>
+         <p align="center">path</p>
+         <img src="img/path.png" alt="path"/>
+         <div align="center"><sub>starting with root and ending with self</sub></div>
```
kbrock, line 123: see above

```suggestion
         <div align="center"><sub>root..self</sub></div>
```

README.md
```diff
@@ -94,100 +94,73 @@ node.children.create :name => 'Stinky'
...
+      </td>
+      <td>
+         <p align="center">children</p>
+         <img src="img/children.png" alt="children"/>
+      </td>
+      <td>
+         <p align="center">siblings</p>
+         <img src="img/siblings.png" alt="siblings"/>
+         <div align="center"><sub>including the record itself</sub></div>
```
kbrock, line 132: footnote

README.md
```diff
@@ -94,100 +94,73 @@ node.children.create :name => 'Stinky'
...
+      </td>
+   </tr>
+   <tr>
+      <td>
+         <p align="center">descendants</p>
+         <img src="img/descendants.png" alt="descendants"/>
+         <div align="center"><sub>returns direct and indirect children</sub></div>
```
kbrock, line 139: is this useful?

README.md
```diff
@@ -94,100 +94,73 @@ node.children.create :name => 'Stinky'
...
+      </td>
+      <td>
+         <p align="center">indirects</p>
+         <img src="img/indirects.png" alt="indirects"/>
+      </td>
+      <td>
+         <p align="center">subtree</p>
+         <img src="img/subtree.png" alt="subtree"/>
+         <div align="center"><sub>returns descendants and itself</sub></div>
```
kbrock, line 148:
```suggestion
         <div align="center"><sub>returns descendants and self</sub></div>
```

README.md
```diff
@@ -208,181 +181,26 @@ The has_ancestry method supports the following options:
...
-
-Thanks to some convenient rails magic, it is even possible to create nodes
-through the children and siblings scopes:
-
-    node.children.create
-    node.siblings.create!
-    TestNode.children_of(node_id).new
-    TestNode.siblings_of(node_id).create
-
-# Selecting nodes by depth
```
kbrock, line 245: thinking this section is useful here

Probably want it somewhere in the yard

README.md
```diff
@@ -208,181 +181,26 @@ The has_ancestry method supports the following options:
...
-    from_depth(depth)       Return nodes starting from a certain depth (node.depth >= depth)
-    after_depth(depth)      Return nodes that are deeper than depth (node.depth > depth)
-
-The depth scopes are also available through calls to descendants,
-descendant_ids, subtree, subtree_ids, path and ancestors. In this case, depth
-values are interpreted relatively. Some examples:
-
-    node.subtree(:to_depth => 2)      Subtree of node, to a depth of node.depth + 2 (self, children and grandchildren)
-    node.subtree.to_depth(5)          Subtree of node to an absolute depth of 5
-    node.descendants(:at_depth => 2)  Descendant of node, at depth node.depth + 2 (grandchildren)
```
kbrock, line 262: lets deprecate?

README.md
```diff
@@ -208,181 +181,26 @@ The has_ancestry method supports the following options:
...
-
-    before_depth(depth)     Return nodes that are less deep than depth (node.depth < depth)
-    to_depth(depth)         Return nodes up to a certain depth (node.depth <= depth)
-    at_depth(depth)         Return nodes that are at depth (node.depth == depth)
-    from_depth(depth)       Return nodes starting from a certain depth (node.depth >= depth)
-    after_depth(depth)      Return nodes that are deeper than depth (node.depth > depth)
-
-The depth scopes are also available through calls to descendants,
-descendant_ids, subtree, subtree_ids, path and ancestors. In this case, depth
-values are interpreted relatively. Some examples:
```
kbrock, line 258: some of the values below are the same as the ones above.

lets deprecate the hash key to_depth (different pr?) it doesn't work as an association. it probably would have worked in rails 3 land, but we're not there anymore.

the ones below just look like examples. not sure if that has value

README.md
```diff
@@ -208,181 +181,26 @@ The has_ancestry method supports the following options:
...
-    node.ancestors(:from_depth => -6, :to_depth => -4)
-    node.path.from_depth(3).to_depth(4)
-    node.descendants(:from_depth => 2, :to_depth => 4)
-    node.subtree.from_depth(10).to_depth(12)
-
-Please note that depth constraints cannot be passed to ancestor_ids and
-path_ids. The reason for this is that both these relations can be fetched
-directly from the ancestry column without performing a database query. It
-would require an entirely different method of applying the depth constraints
-which isn't worth the effort of implementing. You can use
```
kbrock, line 276: these remind me of offset and limit
or slice

would be great if we implemented methods that work on associations
not sure if it is worth the non-backwards compatibility or implementing both

README.md
```diff
@@ -208,181 +181,26 @@ The has_ancestry method supports the following options:
...
-```ruby
-TreeNode.arrange_serializable do |parent, children|
-  {
-     my_id: parent.id,
-     my_children: children
-  }
-end
-```
-
-The result of `arrange_serializable` can easily be serialized to json with
```
kbrock, line 353: lets reduce these comments to what you think is relevant. but maybe we could keep here?

README.md
```diff
@@ -208,181 +181,26 @@ The has_ancestry method supports the following options:
...
-```
-
-You can also pass the order to the `arrange_serializable` method just as you can
-pass it to the `arrange` method:
-
-```
-TreeNode.arrange_serializable(:order => :name)
-```
-
-# Sorting
```
kbrock, line 367: ditto - this is so non-standard

README.md
```diff
@@ -208,181 +181,26 @@ The has_ancestry method supports the following options:
...
-
-If you just want to sort an array of nodes as if you were traversing them in
-preorder, you can use the sort_by_ancestry class method:
-
-```
-TreeNode.sort_by_ancestry(array_of_nodes)
```
kbrock, line 373: dream: `order(:ancestry)`

README.md
```diff
@@ -417,42 +235,7 @@ provide a more detailed explanation:
...
-```
-
-Additionally, if you think something is wrong with your depth cache:
-
-```
->> TreeNode.rebuild_depth_cache!
-```
+</details>
 
 # Running Tests
```
kbrock, line 240: if we want we can have a footnote around appraisals

but this is an ancestry development thing. not sure if that is good or mucking it up

thinking we should have appraisal back in the gemspec as a development dependency?


either way, these docs should only have appraisal install ; appraisal rake test
suggest `wwtd --local` (also not a declared dependency)

README.md
```diff
@@ -475,14 +258,14 @@ This is a variation on the materialised path database pattern.
 It allows Ancestry to fetch any relation (siblings,
```
kbrock, line 258: can we simplify by saying it is materialized path in a string with a `/` delimiter?

README.md
```diff
@@ -82,115 +81,38 @@ parent_id can be set using parent= and parent_id= on a record or by including
...
-node.children.create :name => 'Stinky'
-```
+# Tree Navigation
+
+The node with the large border is the reference node (the node from which the navigation method is invoked.)
+The yellow nodes are those returned by the method.
+
+|                               |                                                     |                                 |
+|:-:                            |:-:                                                  |:-:                              |
+|<b>parent </b>                 |<b>root<sup><a href="#fn1" id="ref1">1</a></sup></b> |<b>ancestors </b>                |
```
kbrock, line 95: can you do `**parent**` to get bold instead?

PR #493, opened 2020-08-01
remove aliases for root, children, childless, siblings, only_children

d-m-u:
minor cleanup. 

these don't buy us anything and i'm not sure why we need them.


kbrock:
I agree with the sentiment. And I was misguided when I introduced them. But not sure I want to add this to "breaking changes"

d-m-u:
i think you're probably okay with me closing this pr but removing the aliases from the table in the readme, does that sound like an compromise?

kbrock:
Yes, please remove the aliases from the readme

PR #495, opened 2020-08-04
sane_ancestor_ids and simplify specs

kbrock:
Introduce sane_ancestor_ids
specs going towards `ancestor_ids`

Prep for ancestor_ids work -- #481

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -116,6 +116,10 @@ module Ancestry
       end
     end
 
+    def sane_ancestor_ids?
+      valid? || errors[self.ancestry_base_class.ancestry_column].blank?
+    end
+
     def ancestors depth_options = {}
       return self.ancestry_base_class.none unless ancestors?
       self.ancestry_base_class.scope_depth(depth_options, depth).ordered_by_ancestry.ancestors_of(self)
--- a/test/concerns/scopes_test.rb
+++ b/test/concerns/scopes_test.rb
@@ -4,7 +4,11 @@ class ScopesTest < ActiveSupport::TestCase
   def test_scopes
     AncestryTestDatabase.with_model :depth => 3, :width => 3 do |model, roots|
       # Roots assertion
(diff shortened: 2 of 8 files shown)


lib/ancestry/instance_methods.rb
```diff
@@ -116,6 +116,10 @@ def ancestry_changed?
       end
     end
 
+    def sane_ancestor_ids?
+      valid? || errors[self.ancestry_base_class.ancestry_column].blank?
+    end
```
kbrock, line 121: this will probably get merged back in when this concept changes. But probably good enough for now

test/concerns/scopes_test.rb
```diff
@@ -55,12 +59,12 @@ def test_order_by
     AncestryTestDatabase.with_model :depth => 3, :width => 3 do |model, roots|
       # not thrilled with this. mac postgres has odd sorting requirements
       if ENV["DB"].to_s =~ /pg/ && RUBY_PLATFORM !~ /x86_64-darwin/
-        expected = model.all.sort_by { |m| [m.ancestry.to_s.gsub('/',''), m.id.to_i] }
+        expected = model.all.sort_by { |m| [m.ancestor_ids.map(&:to_s).join, m.id.to_i] }
```
kbrock, line 62: the goal is to focus on a single format agnostic field

test/concerns/validations_test.rb
```diff
@@ -8,42 +8,29 @@ def test_ancestry_column_validation
         node.send :write_attribute, model.ancestry_column, value
         node.valid?; assert node.errors[model.ancestry_column].blank?
       end
-      ['1/3/', '/2/3', 'a', 'a/b', '-34', '/54'].each do |value|
```
kbrock, line 11: we used to focus that the format of this field was serialized correctly.

The new functionality parses these values just fine.
And since we are moving the serialization to a single method - the chances of failure is much lower than when this field was formed all over the code.

PR #496, opened 2020-08-04
Update changelog date for 3.1.0

CamilleDrapier:
## Why

I was a bit surprised to see an update "popping up" from dependabot with the last changed said to be in 2018... so I started looking at what triggered this update to trigger before realizing it was just probably a simple copy-typo.

## What

Update the date in the chagelog; I used the "UTC" date of the release for this.


kbrock:
Thanks Camille.

PR #497, opened 2020-08-04
[wip] pg optimization

d-m-u:
add postgres optimizer with regexp_replace 

all from https://github.com/stefankroes/ancestry/pull/464, wanted to get some traction on this 

@kbrock could you give this another look for me please and let me know what else I need to do here?

### DIFF
--- a/README.md
+++ b/README.md
@@ -468,11 +468,6 @@ appraisal rake test
 appraisal sqlite3-ar-50 rake test
 ```
 
-# Optimization
-
-Ancestry provides a way to optimize for specific database. Currently, there is a optimizer for Postgres.
-To enable it, you need to include: `require 'ancestry/optimizers/pg'` after `require 'ancestry'`.
-
 # Internals
 
 Ancestry stores a path from the root to the parent for every node.
--- a/README.md
+++ b/README.md
@@ -468,6 +468,11 @@ appraisal rake test
 appraisal sqlite3-ar-50 rake test
 ```
 
+# Optimization
+
+Ancestry provides a way to optimize for specific database. Currently, there is a optimizer for Postgres.
+To enable it, you need to include: `require 'ancestry/optimizers/pg'` after `require 'ancestry'`.
+
 # Internals
 
 Ancestry stores a path from the root to the parent for every node.
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -4,7 +4,7 @@ module Ancestry
       # Check options
       raise Ancestry::AncestryException.new("Options for has_ancestry must be in a hash.") unless options.is_a? Hash
       options.each do |key, value|
-        unless [:ancestry_column, :orphan_strategy, :cache_depth, :depth_cache_column, :touch, :counter_cache, :primary_key_format].include? key
+        unless [:ancestry_column, :orphan_strategy, :cache_depth, :depth_cache_column, :touch, :counter_cache, :primary_key_format, :update_strategy].include? key
           raise Ancestry::AncestryException.new("Unknown option for has_ancestry: #{key.inspect} => #{value.inspect}.")
         end
       end
@@ -24,13 +24,14 @@ module Ancestry
       # Include instance methods
       include Ancestry::InstanceMethods
 
-      include Ancestry.optimizer if Ancestry.optimizer
-
       # Include dynamic class methods
       extend Ancestry::ClassMethods
 
       validates_format_of self.ancestry_column, :with => derive_ancestry_pattern(options[:primary_key_format]), :allow_nil => true
       extend Ancestry::MaterializedPath
+
+      update_strategy = options[:update_strategy] || Ancestry.default_update_strategy
+      include Ancestry::MaterializedPathPg if options[:update_strategy] == :sql && ActiveRecord::Base.connection.adapter_name.downcase == "postgresql"
 
       # Create orphan strategy accessor and set to option or default (writer comes from DynamicClassMethods)
       cattr_reader :orphan_strategy
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -24,6 +24,10 @@ module Ancestry
       # Include instance methods
       include Ancestry::InstanceMethods
 
+      if Ancestry.optimizer
+        include Ancestry.optimizer
+      end
+
       # Include dynamic class methods
       extend Ancestry::ClassMethods
 
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -24,9 +24,7 @@ module Ancestry
       # Include instance methods
       include Ancestry::InstanceMethods
 
-      if Ancestry.optimizer
-        include Ancestry.optimizer
-      end
+      include Ancestry.optimizer if Ancestry.optimizer
 
       # Include dynamic class methods
       extend Ancestry::ClassMethods
--- a/lib/ancestry/materialized_path_pg.rb
+++ b/lib/ancestry/materialized_path_pg.rb
@@ -1,13 +1,12 @@
 module Ancestry
-  module Optimizers
-    module Pg
+  module MaterializedPathPg
       # Update descendants with new ancestry (before save)
       def update_descendants_with_new_ancestry
         # If enabled and node is existing and ancestry was updated and the new ancestry is sane ...
         if !ancestry_callbacks_disabled? && !new_record? && ancestry_changed? && sane_ancestry?
           ancestry_column = ancestry_base_class.ancestry_column
-          old_ancestry = child_ancestry
-          new_ancestry = ancestors? ? "#{read_attribute ancestry_column}/#{id}" : id.to_s
+          old_ancestry = path_ids_in_database.join(Ancestry::MaterializedPath::ANCESTRY_DELIMITER)
+          new_ancestry = path_ids.join(Ancestry::MaterializedPath::ANCESTRY_DELIMITER)
           update_clause = [
             "#{ancestry_column} = regexp_replace(#{ancestry_column}, '^#{old_ancestry}', '#{new_ancestry}')"
           ]
@@ -18,10 +17,7 @@ module Ancestry
           end
 
           unscoped_descendants.update_all update_clause.join(', ')
-        end
       end
     end
   end
 end
-
-Ancestry.optimizer = Ancestry::Optimizers::Pg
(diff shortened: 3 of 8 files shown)


lib/ancestry/optimizers/pg.rb
```diff
@@ -0,0 +1,27 @@
+module Ancestry
+  module Optimizers
```
kbrock, line 2: I'd prefer to avoid the whole optimizer thing. This is simply how we do materialized path for postgres.
And this implementation is specific to postgres

just call it: `Ancestry::MaterializedPathPg`

lib/ancestry/has_ancestry.rb
```diff
@@ -30,6 +30,8 @@ def has_ancestry options = {}
       validates_format_of self.ancestry_column, :with => derive_ancestry_pattern(options[:primary_key_format]), :allow_nil => true
       extend Ancestry::MaterializedPath
 
+      include Ancestry.optimizer if Ancestry.optimizer
```
kbrock, line 33: it may change in the future, but for now lets hardcode this thing

``
include Ancestry::MaterializedPathPg connection.adapter_name.downcase == "postgresql"
```

and for a test maybe make sure it has the correct ancestor.

```
if `ENV["DB"].to_s =~ /pg/i`
  model.ancestors.map(&:to_s).include?("Ancestry::MaterializedPathPg")
end
```

File: lib/ancestry/optimizers/pg.rb [hunk 3]
```diff
@@ -0,0 +1,27 @@
...
+          update_clause = [
+            "#{ancestry_column} = regexp_replace(#{ancestry_column}, '^#{old_ancestry}', '#{new_ancestry}')"
+          ]
+
+          if ancestry_base_class.respond_to?(:depth_cache_column) && respond_to?(ancestry_base_class.depth_cache_column)
+            depth_cache_column = ancestry_base_class.depth_cache_column.to_s
+            update_clause << "#{depth_cache_column} = length(regexp_replace(regexp_replace(ancestry, '^#{old_ancestry}', '#{new_ancestry}'), '\\d', '', 'g')) + 1"
+          end
+
+          unscoped_descendants.update_all update_clause.join(', ')
```
kbrock, line 20: 1. if we're doing this in sql, then locally cached models will be unchanged (and wrong).
2. update hooks are not called.

You will notice that `depth_cache_column` is reimplemented here. (we need a test that this is updated for children)

Another example: A model that stores the names of all the ancestors will have something like a `before_save` hook that changes the ancestor_names field. The update statement will need to be updated for these records.

All in all, this has similar tradeoffs as `destroy` vs `delete`.

Wonder if there is a mechanism that models can use to enhance that update clause.
Thinking that depth_cache_column needs to also use this mechanism rather than modifying this method.

When I finally introduce  `ancestor_ids` array column, I'll run into a similar optimization situation.

test/environment.rb
```diff
@@ -47,8 +47,9 @@ def self.setup
       File.expand_path('../database.ci.yml', __FILE__)
     end
 
+    use_optimizer
```
kbrock, line 50: yea, think this functionality should be automatic performed in `has_ancestry` - either triggered by a parameter or some environmental check.

README.md
```diff
@@ -468,6 +468,11 @@ appraisal rake test
 appraisal sqlite3-ar-50 rake test
 ```
 
+# Optimization
```
kbrock, line 471: remove

lib/ancestry/has_ancestry.rb
d-m-u: hmm, [you sure about that line?](https://github.com/stefankroes/ancestry/blob/master/test/concerns/db_test.rb#L7)

lib/ancestry/materialized_path_pg.rb
```diff
@@ -0,0 +1,25 @@
...
+          update_clause = [
+            "#{ancestry_column} = regexp_replace(#{ancestry_column}, '^#{old_ancestry}', '#{new_ancestry}')"
+          ]
+
+          if ancestry_base_class.respond_to?(:depth_cache_column) && respond_to?(ancestry_base_class.depth_cache_column)
+            depth_cache_column = ancestry_base_class.depth_cache_column.to_s
+            update_clause << "#{depth_cache_column} = length(regexp_replace(regexp_replace(ancestry, '^#{old_ancestry}', '#{new_ancestry}'), '\\d', '', 'g')) + 1"
+          end
+
+          unscoped_descendants.update_all update_clause.join(', ')
```
d-m-u, line 19: I didn't address the feedback on this line yet, will get around to it later

kbrock: ok, so there was the opt-in thing.

And there was the concern that locally cached objects would be stale.

test/environment.rb
kbrock: delete all changes to this file

lib/ancestry/optimizer.rb
```diff
@@ -0,0 +1,9 @@
+module Ancestry
```
kbrock, line 1: drop this class

lib/ancestry/materialized_path_pg.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+    end
+  end
+end
+
+Ancestry.optimizer = Ancestry::MaterializedPathPg
```
kbrock, line 25: yea, make optimizer go away

README.md
```diff
@@ -468,6 +468,9 @@ appraisal rake test
 appraisal sqlite3-ar-50 rake test
 ```
 
+Ancestry provides a Postgres optimizer. To enable it, you need to include:
+`require 'ancestry/materialized_path_pg'` after `require 'ancestry'`.
```
kbrock, line 472: no need for these comments

lib/ancestry/has_ancestry.rb
```diff
@@ -30,6 +30,8 @@ def has_ancestry options = {}
       validates_format_of self.ancestry_column, :with => derive_ancestry_pattern(options[:primary_key_format]), :allow_nil => true
       extend Ancestry::MaterializedPath
 
+      include Ancestry::MaterializedPathPg if Ancestry.optimizer
```
kbrock, line 33: Yea, I'm still leaning towards this.
Please share where I flubbed it

```suggestion
      include Ancestry::MaterializedPathPg if connection.adapter_name.downcase == "postgresql"
```

Think this is what we do in other places.
If connection can not be accessed, it may be `class.connection` or something like that? Pretty sure `self` is the class coming in.

kbrock: If you think optimization should be opt in, then you add an opt-in parameter (`:sql_updates =>true`)

I do feel it is model specific than database specific.
Just like `:dependent => :destroy` and `:dependent => :delete`

lib/ancestry/materialized_path_pg.rb
d-m-u: yeah, i concur about your opt-in feedback, and i'll get to the caching issues on monday

lib/ancestry/has_ancestry.rb
d-m-u: not marking this as resolved because i haven't gotten around to the feedback about the test either

kbrock:
I want everyone to get this performance benefit. 

If we include this patch for everyone (running postgres) then the standard test suite will test both code paths: the optimization postgres route and the non-optimized route. The first running in postgres and the second running for mysql and sqlite tests

But I do worry about the user who gets this optimization and doesn't realize that after update is not called for child records unless specifically requested.

I like the idea of adding a hook so we can efficiently update records with custom after update hooks. Much more efficient and faster. But potentially clever and hidden.

lib/ancestry/has_ancestry.rb
```diff
@@ -30,6 +30,8 @@ def has_ancestry options = {}
       validates_format_of self.ancestry_column, :with => derive_ancestry_pattern(options[:primary_key_format]), :allow_nil => true
       extend Ancestry::MaterializedPath
 
+      include Ancestry::MaterializedPathPg if ActiveRecord::Base.connection.adapter_name.downcase == "postgresql"
```
kbrock, line 33: introduce `update_strategy` that will allow us to opt into this behavior

please keep the postgresql check here
but add a check to make sure update strategy is sql



for those who want this everywhere, add a module level variable with default update strategy

kbrock, line 33: introduce `update_strategy` that will allow us to opt into this behavior

possibly introduce Ancestry.default_update_strategy defaulted to `:ruby` 

please keep the postgresql check here
but add a check to make sure update strategy is sql

```ruby
class Polygon
  @@sides = 10
  def self.sides
    @@sides
  end
end

puts Polygon.sides # => 10
```

how can we test this? I'm thinking that we should set this to sql in our test suite
so it will become opt out

lib/ancestry/has_ancestry.rb
```diff
@@ -30,6 +30,13 @@ def has_ancestry options = {}
       validates_format_of self.ancestry_column, :with => derive_ancestry_pattern(options[:primary_key_format]), :allow_nil => true
       extend Ancestry::MaterializedPath
 
+      @@default_update_strategy = :ruby
```
kbrock, line 33: I think we want to turn this on for all tests in `environment.rb`


When/if we have optimizers for all databases we may decide to introduce a ruby test, but in the meantime, mysql and sqlite will test these values

lib/ancestry/has_ancestry.rb
```diff
@@ -30,6 +30,13 @@ def has_ancestry options = {}
       validates_format_of self.ancestry_column, :with => derive_ancestry_pattern(options[:primary_key_format]), :allow_nil => true
       extend Ancestry::MaterializedPath
 
+      @@default_update_strategy = :sql
+      def self.default_update_strategy
+        @@default_update_strategy
+      end
```
kbrock, line 36: wait, why did you put this here?
I thought it would be in Ancestry module and usable as a global for all

That way in environment.rb we could set this to sql for all tests.

I liked how you had this as :ruby before.

tests: use sql
default for everyone else: sql

lib/ancestry/has_ancestry.rb
```diff
@@ -30,6 +30,13 @@ def has_ancestry options = {}
...
+
+      include Ancestry::MaterializedPathPg if options[:update_strategy] == :sql && ActiveRecord::Base.connection.adapter_name.downcase == "postgresql"
```
kbrock, line 38: perfect

lib/ancestry/has_ancestry.rb
kbrock: sorry I was not clear here

kbrock:
merged via command line - thanks

PR #501, opened 2020-08-10
remove code for old activerecord versions

d-m-u:
do we still need any of this? 

i think we're locked to >= 4.2

the monkey patch and unscoped_where if clause are < 4, the primary key as int thing is < 4.2 

really my goal is to get the test coverage to 100 and clearly the way to do that is taking code out
/s

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -207,31 +207,15 @@ module Ancestry
     end
 
     def unscoped_where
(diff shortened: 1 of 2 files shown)


kbrock:
I'll put out a 3.x release and then I'll merge this - which forces us up to 4.x - and then the expectation that we are doing tons of stuff here like fixing that blasted warning in rails 6.1

thinking about dropping 5.0 as well. 5.1 fixes some active record stuff that would be nice to use.

PR #507, opened 2020-09-18
Remove postgresql check for :sql update strategy

greis:
I was running into an issue when running `rake asset:precompile` in a CI environment since it didn't have any postgres database running. It was failing because `ActiveRecord::Base.connection` tries to connect to the database.

`update_strategy: :sql` is an opt-in option so it's probably not necessary to do the postgres check.

What are your thoughts?

### DIFF
--- a/test/environment.rb
+++ b/test/environment.rb
@@ -55,7 +55,7 @@ class AncestryTestDatabase
 
       # This only affects postgres
       # the :ruby code path will get tested in mysql and sqlite3
-      Ancestry.default_update_strategy = :sql if ActiveRecord::Base.connection.adapter_name.downcase == "postgresql"
+      Ancestry.default_update_strategy = :sql if db_type == "pg"
 
     rescue => err
       if ENV["CI"]
--- a/test/environment.rb
+++ b/test/environment.rb
@@ -47,15 +47,16 @@ class AncestryTestDatabase
       File.expand_path('../database.ci.yml', __FILE__)
     end
 
-    # This only affects postgres
-    # the :ruby code path will get tested in mysql and sqlite3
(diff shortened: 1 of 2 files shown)


kbrock:
@greis thanks for finding this. I get frustrated when I declare a class and it tries to load the database.

The concern is that a mysql database user requests optimizations but there is none for them.
And future expansion here into other optimizations and possible string concatenation functions to fix our ruby only nature. 
but you pointed out a problem and we can't discover the database type at class creation / module include time.

A user can pass in their database type in if they want to take advantage of this optimization. And all new users will probably want to just turn this on for all models in an initializer (what we do for tests).

/cc @d-m-u

greis:
@kbrock I see. So technically with the previous code a mysql user would request optimizations but they wouldn't get any warning or error, right?

test/environment.rb
```diff
@@ -47,15 +47,16 @@ def self.setup
...
-
     # Setup database connection
     config = YAML.load_file(filename)[db_type]
     ActiveRecord::Base.establish_connection config
     begin
       ActiveRecord::Base.connection
+
+      # This only affects postgres
+      # the :ruby code path will get tested in mysql and sqlite3
+      Ancestry.default_update_strategy = :sql if ActiveRecord::Base.connection.adapter_name.downcase == "postgresql"
```
kbrock, line 58: Think we already know the db type here, lets do:

```suggestion
      # the :ruby code path will get tested in mysql and sqlite3
      Ancestry.default_update_strategy = :sql if dbtype == "pg"
```

kbrock (commented):
just the one minor nit and then lets get this in

kbrock:
thank you @greis

Issue #510, opened 2020-09-23
Unable to update gem to 3.2.0


mjerem34:
Hello, it is impossible for me to update the version of the gem
I have this error on CircleCi and Heroku : 
```
rake aborted!
LoadError: cannot load such file -- /home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/ancestry-3.2.0/lib/ancestry/materialized_path_pg
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `require'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `block in require_with_bootsnap_lfi'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/loaded_features_index.rb:92:in `register'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:22:in `require_with_bootsnap_lfi'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:31:in `require'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/activesupport-6.0.2.2/lib/active_support/dependencies.rb:325:in `block in require'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/activesupport-6.0.2.2/lib/active_support/dependencies.rb:291:in `load_dependency'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/activesupport-6.0.2.2/lib/active_support/dependencies.rb:325:in `require'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:49:in `require_relative'
/home/circleci/repo/vendor/bundle/ruby/2.7.0/gems/ancestry-3.2.0/lib/ancestry.rb:7:in `<main>'
[...]

PR #511, opened 2020-09-23
Remove require for materialized_path_pg

techpeace:
The require for `materialized_path_pg` is breaking v3.2.0. This is a proposed fix for #510.


techpeace:
Looks like there's a better fix in #512

kbrock:
thank you for your help. I'll go with the other fix

PR #512, opened 2020-09-23
Bundled gem is missing files

HectorMF:
Fix for #510 

The gem found on rubygems.org is missing **materialized_path_pg.rb** as well as the **locales** folder because they are not included in the the gemspec files list. 

```
  s.files = [
    'ancestry.gemspec',
    'init.rb',
    'install.rb',
    'lib/ancestry.rb',
    'lib/ancestry/has_ancestry.rb',
    'lib/ancestry/exceptions.rb',
    'lib/ancestry/class_methods.rb',
    'lib/ancestry/instance_methods.rb',
    'lib/ancestry/materialized_path.rb',
    'lib/ancestry/version.rb',
    'MIT-LICENSE',
    'README.md'
  ]
```

To prevent this from happening in future releases I've standardized the .gemspec, requiring the lib folder by default. I've removed the init.rb and install.rb files which are not needed, and added the CHANGELOG as well.

### DIFF
--- a/ancestry.gemspec
+++ b/ancestry.gemspec
@@ -14,6 +14,9 @@ Gem::Specification.new do |s|
   of (sub)tree into hashes and different strategies for dealing with orphaned
   records.
 EOF
+
+  s.post_install_message = "Thank you for installing Ancestry. You can visit http://github.com/stefankroes/ancestry to read the documentation."
+
   s.metadata = {
     "homepage_uri" => "https://github.com/stefankroes/ancestry",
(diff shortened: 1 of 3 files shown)


kbrock:
thanks. I'll follow up with a tweak to use git instead of ls

kbrock:
take that back. Not thrilled about having a wildcard for every file in the directory, but there sure is a lot of files in there for testing that we do not want to include.

I like your option the best so far

PR #514, opened 2020-10-08
drop rails 4.x

kbrock:
going towards using the attributes api

still supporting rails 4.x in ancestry 3.2


d-m-u:
looks great, could you also remove the 4.2 stuff from the Appraisals file?
grep's giving me three lines in particular: 

```ruby
Appraisals:1:%w(4.2.10 5.0.7 5.1.7 5.2.3 6.0.0).each do |ar_version|
Appraisals:10:    # rails 4.2 supports both ( but travis complains with 4.2 and mysql2)
Appraisals:11:    if ar_version < "4.2"
```

d-m-u (approved):
approving contingent on resolution of the minor nit in the previous comment

PR #518, opened 2020-10-18
No need to send a public method "Object#extend"

amatsuda:
`Object#extend` has been defined as a public method since the last century (I don't exactly know when, but it already was a public method even on version 1.0.0).
So we don't need to call it via `send`.


kbrock:
thanks

PR #522, opened 2020-12-16
Fixes Rails 6 Scoping Deprecation Warnings

cmr119:
We recently updated our app to rails 6.0 and noticed one of the deprecation warnings documented in #442.  Specifically:
```
DEPRECATION WARNING: Class level methods will no longer inherit scoping from `create!` in Rails 6.1. To continue using the scoped relation, pass it into the block directly. To instead access the full set of models, as Rails 6.1 will, use `<model name>.default_scoped`.
```

After this small change the gem's test passed for me locally, and we have a fairly thorough suite of specs around the ancestry functionality that also passed.  So, I figured it was worth opening a PR for any additional feedback.

Thanks for all your work supporting this gem!

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -217,7 +217,7 @@ module Ancestry
     end
 
     def unscoped_where
-      yield self.ancestry_base_class.unscope(:where)
+      yield self.ancestry_base_class.default_scoped.unscope(:where)
     end
 
     ANCESTRY_UNCAST_TYPES = [:string, :uuid, :text].freeze


kbrock:
kicking to run the tests again.

I'm a little nervous using default scopes because I think I remember some people specifically wanting the scopes to come into play in some STI scenarios

BenHand:
What is the status of this? Converting my application to 6.1 soon and hoping to remove the deprecation warnings before doing so.

kbrock:
@brendon Have you had a chance to test with this?

The goal of this change (or at least one of them) was to ensure that the order came back in the same order as other queries so we did not end up with deadlocks.

Will these changes still meet your needs?

https://github.com/stefankroes/ancestry/pull/348

brendon:
Hi @kbrock, we fixed this in `acts_as_list` in a much simpler way from memory. Take a look there. I can’t quite remember how we did it but I’ll look into it after the holidays :)

brendon:
Hi @kbrock, in following up, this is exactly what we did with `acts_as_list`. The key was still doing the `unscope` afterward still as `default_scoped` returns a scope with defaults even if called on a model class itself.

kbrock:
thanks @brendon

kbrock:
thanks @cmr119 this is a long standing issue - appreciate your help

PR #525, opened 2021-01-04
add: rails 6.1 and ruby 3

jellymann:
Adding Rails 6.1 and Ruby 3 to the test matrix, and fixing issues causing failures in Ruby 3.


jellymann:
There seems to be a breaking change with the way ordering works in ActiveRecord 6.1 with postgres, not sure what the issue is.

jellymann:
Also, I need to find a way for Ruby 3 to _only_ run for Rails 6.1, since the older versions of Rails just break. I tried putting a check in the Appraisals file but it didn't do the trick.

kbrock:
thank you - we need to get rid of some older rails versions to get this green again

kbrock:
the ordering is tricky because different encodings have different sort orders - specifically to us, some include '/' in the sorting and others do not.

We could limit the ruby part of the matrix, but removing old versions seems to be the best bet right now

kbrock:
thank you @jellymann I resolved the conflict manually and merged your changes. (I did trim just a little bit out of your change but it is basically the same thing)

Left attribution, but didn't know how to make it know that this PR is the one that I merged. sorry this shows red instead of purple

jellymann:
Thanks @kbrock no problem at all, glad it got in! When do you expect you'll release a new version with this change in it?

kapso:
@kbrock curious about this release as well, thanks.

kbrock:
@kapso thank you for your gentle reminder

jellymann:
@kbrock anything you need help with? I've got some time this week after Easter

PR #526, opened 2021-01-12
drop support for 5.0 and 5.1

d-m-u:
i don't think we really need to keep supporting these

the only thing i didn't change was in `instance_methods.rb`:
```
    def decrease_parent_counter_cache
      # @_trigger_destroy_callback comes from activerecord, which makes sure only once decrement when concurrent deletion.
      # but @_trigger_destroy_callback began after rails@5.1.0.alpha.
      # https://github.com/rails/rails/blob/v5.2.0/activerecord/lib/active_record/persistence.rb#L340
      # https://github.com/rails/rails/pull/14735
      # https://github.com/rails/rails/pull/27248
      return if defined?(@_trigger_destroy_callback) && !@_trigger_destroy_callback
      return if ancestry_callbacks_disabled?
```
if you like this, could you take a look at that file and tell me what else i can remove please? 

thanks


lib/ancestry/instance_methods.rb
```diff
@@ -107,13 +102,9 @@ def has_parent?
 
     def ancestry_changed?
       column = self.ancestry_base_class.ancestry_column.to_s
-      if ActiveRecord::VERSION::STRING >= '5.1.0'
```
kbrock, line 110: maybe we want to add a rails 6 fix here?
per the comment below
that is probably a different pr

Appraisals
```diff
@@ -1,4 +1,4 @@
-%w(5.0.7 5.1.7 5.2.3 6.0.0).each do |ar_version|
+%w(5.2.3 6.0.0).each do |ar_version|
```
kbrock, line 1: do we want to set the minimum version in the gemspec

does this mean we want to delete gemfiles as well?

kbrock:
now that we only do 5.2 and above, we want to use the standard will_change/saved ones instead

this is a breaking change and we'll probably introduce when this is rewritten to dynamically create these methods

PR #527, opened 2021-01-30
Fix order_by_ancestry_and for ActiveRecord 6.1

smoyth:
Restored the second arg to `reorder`


kbrock:
how the heck was this green?

smoyth:
Thank you!

PR #535, opened 2021-04-16
Fix ordered_by_ancestry for oracle-db

lual:
An empty string is treated as NULL in Oracle. 
`COALESCE(NULL,'')` is treated as NULL. 
So Ordering ascending, deliver NULL-Values at the rear.


kbrock:
@lual could you run with the patch from #541 and verify it meets your needs?

I'm pretty sure the postgresql solution will meet your needs. Wrapping in the ascend allows reverse ordering not to blow up.

kbrock:
Better yet, would you make the changes I made over there and put them into your pr? that way you could get credit

kbrock:
hello @lual 

Thank you for posting.

postgres and probably quite a few ANSI SQL databases support the nulls first option.

I would not worry about the sorting order test failure - we're having issues with the postgres locale used by travis and github

kbrock:
I'll give you a few days, otherwise I'll just merge that other PR and hopefully that will meet your needs

kbrock:
monkey patch in #535 should hold you over until the next version of rails is released

PR #536, opened 2021-04-16
Missing parent

### LINKED ISSUE #523: Suggestion:  parent should return nil instead of ActiveRecordNotFound exception.
Presently, when the `ancestry` field contains an ID, but the parent object has been deleted from the database, calling the `parent` method results in an exception. 

```ruby
user.parent = user2
user2.delete
user.parent
> ActiveRecordNotFound
```

In contrast, ActiveRecord returns nil for a `belongs_to` relation, when the foreign key has a value, but the record does not exist. 

For this example, a User "belongs_to" a Team.

```ruby
user.update_column :team_id, 123456
user.team
> nil
```
Please consider making the result of the `parent` method consistent with the handling of `belongs_to` in ActiveRecord. 

Thanks!

kbrock:
resolves #523

if an invalid value is in the `parent_id`, `parent` will return nil not throw error

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 gemspec
 
-gem "activerecord", '~> 5.2.4.5'
+gem "activerecord", "~> 6.0.3"
 gem "coveralls", require: false
 gem "mysql2"
 gem "pg"
--- a/test/concerns/relations_test.rb
+++ b/test/concerns/relations_test.rb
@@ -1,0 +1,37 @@
+require_relative '../environment'
+
+class RelationsTest < ActiveSupport::TestCase
+  def test_root_found
+    AncestryTestDatabase.with_model do |model|
+      parent = model.create
+      child = model.create!(:ancestor_ids => [parent.id])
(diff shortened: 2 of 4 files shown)


d-m-u (approved):
do you want to merge https://github.com/stefankroes/ancestry/pull/538 before this?

test/concerns/relations_test.rb
```diff
@@ -0,0 +1,37 @@
+require_relative '../environment'
+require 'byebug'
```
d-m-u, line 2: this line, do you need?

kbrock: JUST deleted before I saw this - thanks

Gemfile
```diff
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 gemspec
 
-gem "activerecord", '~> 5.2.4.5'
+gem "activerecord", "~> 6.0.3"
```
kbrock, line 5: of note, for these we want to limit to 3 digits. the last digit is allowed to change.
So pegging to 5.2.4.5 will allow 5.2.4.6 but not 5.2.5.

PR #538, opened 2021-04-16
[work in progress] drop active record five two

d-m-u:
hi, I think we also want to do this sometime.


kbrock:
5.2 supports ruby 2.2 - it still is under security release and has a little life left in it.
I suppose we could say that 5.2 is still supported in the non-head. That would allow us to drop a few of the monkey patches we needed. Think most of the monkey patches were for 5.1

kbrock:
oh dear - per https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html

looks like ruby 2.5 is the common denominator for these.

PR #541, opened 2021-05-14
oracle use NULLS first

kbrock:
the order of nulls relative to columns with values is setup as
a default value in databases, which can be changed. The default for
postgres is to have nulls last. Looks like oracle has the same.

This PR has oracle enhanced driver follow suite as postgres.

Postgres and rails 6.0 and earlier had issues with the null first, so this is
feature limited to 6.1 and later

Alternative to https://github.com/stefankroes/ancestry/pull/535

@d-m-u this is another example of where the defaults of the database can really affect the way this gem works.
Add this to the list of configuration options that we would like to have. So someone can say that nulls first is part of the database and they don't need to add this funky syntax

### DIFF
--- a/lib/ancestry/materialized_path.rb
+++ b/lib/ancestry/materialized_path.rb
@@ -67,7 +67,7 @@ module Ancestry
     def ordered_by_ancestry(order = nil)
       if %w(mysql mysql2 sqlite sqlite3).include?(connection.adapter_name.downcase)
         reorder(arel_table[ancestry_column], order)
-      elsif %w(postgresql).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::STRING >= "6.1"
+      elsif %w(postgresql oracleenhanced).include?(connection.adapter_name.downcase) && ActiveRecord::VERSION::STRING >= "6.1"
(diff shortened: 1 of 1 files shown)


lual:
@kbrock 
This patch (as alternative to #535) doesn't work for oracle-enhanced. I've tried that.
It seems that the `nulls_first` -method is implemented for postgresql only. see https://github.com/rails/rails/pull/38131

kbrock:
@lual I have fixed this in rails core. If you are not using mysql, then you could probably just monkey patch your code - of course putting an activerecord version limiter so you know to undo it in the future.

please see https://github.com/rails/rails/pull/42245

With this patch, will the code now work?

kbrock:
kicking and crossing fingers

kbrock:
@lual I think this monkey patch should fix your issue.
This fixes oracle until the next version of rails is released. It will break mysql in some edge cases.

```ruby
module Ancestry
  module ToSqlPatch
    def visit_Arel_Nodes_NullsFirst(o, collector)
      visit o.expr, collector
      collector << " NULLS FIRST"
    end

    def visit_Arel_Nodes_NullsLast(o, collector)
      visit o.expr, collector
      collector << " NULLS LAST"
    end
  end
end
Arel::Visitors::ToSql.extend(Ancestry::ToSqlPatch)
```

PR #542, opened 2021-05-20
change the valid check to not only run on the entire object

d-m-u:
Per https://github.com/ManageIQ/manageiq/pull/21226#discussion_r636137634


kbrock:
kicking

kbrock:
maybe you can add a test around updating a model with invalid data in it or something?
not sure the best way to test this use case

PR #544, opened 2021-05-21
upgrade mysql

kbrock:
getting some strange mysql failures.

upgrading mysql to try and fix


kbrock:
this is red, but no more red than master right now.
it is fixing the ssl error

PR #545, opened 2021-05-21
remove coveralls

kbrock:
getting some errors with coveralls

not really paying attention to these metrics. removing this to simplify the gem dependencies

/cc @Fryguy do you have a major opinion on this change?


Fryguy:
I'm fine with dropping it - in other projects I've switched to CodeClimate's coverage.

Issue #549, opened 2021-05-31
STI does not work with counter cache


mattvague:
It looks like the counter cache feature of ancestry does not work when using different STI subclasses in your tree. For example, let's say you have a setup like

```ruby
class Node
  has_ancestry counter_cache: true
end

class SomeNodeType < Node
end

class SomeOtherNodeType < Node
end

root = SomeNodeType.create
edge = SomeOtherNodeType.create(parent: SomeNodeType)
```

Calling `root.children_count` will return 0 where it should have returned 1. I dug into it and I believe it's due to calling `self.class.increment_counter` instead of `self.ancestry_base_class.increment_counter` [here](https://github.com/stefankroes/ancestry/blob/17c73ba9cc5b1d8b91f9021344c781f85e1bc64c/lib/ancestry/instance_methods.rb#L65).

@kbrock Is this intentional or is this something you'd accept a PR to change?

Thanks!

mattvague:
My proposed fix is here: https://github.com/goodproblems/ancestry/commit/e17cf5e28b224a7b9dacc291eff622a4951c7b9f

kbrock:
This is great. PR would be gladly accepted. Do you need help with a spec?

kbrock:
Do you think https://github.com/stefankroes/ancestry/blob/17c73ba9cc5b1d8b91f9021344c781f85e1bc64c/lib/ancestry/instance_methods.rb#L74 without the same guard in the increment is a problem as well? (Just noticed this while looking at your code)

mattvague:
>This is great. PR would be gladly accepted. Do you need help with a spec?

Yeah that would be great actually as I was having some trouble getting ancestry suite running due to mysql gem issues on my M1 Mac. I added a test that should _in theory_ work, but let me know what you think. PR is here: https://github.com/stefankroes/ancestry/pull/550

> Do you think ... without the same guard in the increment is a problem as well? (Just noticed this while looking at your code)

That's a good question, I'm not familiar with that part of rails or ancestry so I'd have to look into it more. Spontaneously I would say that I can't see a reason to have that on decrement but not increment. Probably best to see if you can write a failing test that replicates the issue?

mattvague:
@kbrock One more thing: there were a few other things that I wanted to discuss / propose for ancestry but I think issues are not the right place. Any idea what it would take to get GitHub Discussions enabled?

mattvague:
[...]

PR #550, opened 2021-06-01
Fix issue with counter cache and STI

mattvague:
Addresses https://github.com/stefankroes/ancestry/issues/549

### DIFF
--- a/test/concerns/sti_support_test.rb
+++ b/test/concerns/sti_support_test.rb
@@ -35,4 +35,30 @@ class StiSupportTest < ActiveSupport::TestCase
       Object.send :remove_const, 'SubclassWithAncestry'
     end
   end
+
+  def test_sti_support_for_counter_cache
+    AncestryTestDatabase.with_model :counter_cache => true, :extra_columns => {:type => :string} do |model|
+      # NOTE had to use subclasses other than Subclass1/Subclass2 from above
+      # due to (I think) Rails caching those STI classes and that not getting
+      # reset between specs
+
+      Object.const_set 'Subclass3', Class.new(model)
+      Object.const_set 'Subclass4', Class.new(model)
+
+      node1 = Subclass3.create!
+      node2 = Subclass4.create! :parent => node1
+      node3 = Subclass3.create! :parent => node1
+      node4 = Subclass4.create! :parent => node3
+      node5 = Subclass3.create! :parent => node4
+
+      assert_equal 2, node1.reload.children_count
+      assert_equal nil, node2.reload.children_count
+      assert_equal 1, node3.reload.children_count
+      assert_equal 1, node4.reload.children_count
+      assert_equal nil, node5.reload.children_count
+
+      Object.send :remove_const, 'Subclass3'
+      Object.send :remove_const, 'Subclass4'
+    end
+  end
 end
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -62,7 +62,7 @@ module Ancestry
 
     # Counter Cache
     def increase_parent_counter_cache
-      self.class.increment_counter _counter_cache_column, parent_id
+      self.ancestry_base_class.increment_counter _counter_cache_column, parent_id
     end
 
     def decrease_parent_counter_cache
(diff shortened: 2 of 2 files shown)


kbrock:
please get rid of the WIP comment

kbrock:
I fixed the STI classes defined in https://github.com/stefankroes/ancestry/pull/553

Even with these quick changes to your tests, the tests do not properly work for me.
Could you take a quick peek?

thanks

mattvague:
@kbrock Hey sorry about that, busy week. I'm not surprised the tests don't work as I never got them actually running haha. I'm going to try again now and I'll report back what happens with MySQL on my M1.

mattvague:
@kbrock Could you try running the specs? Don't think I can

kbrock:
> @kbrock Could you try running the specs? Don't think I can

not sure what is up with this. have never seen this before

kbrock:
yea, this is showing the same error I was getting locally. returning `nil` instead of `2`

mattvague:
@kbrock It says

> First-time contributors need a maintainer to approve running workflows. Learn more.

https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

mattvague:
>yea, this is showing the same error I was getting locally. returning nil instead of 2

Anyways, think I fixed it!

test/concerns/sti_support_test.rb
```diff
@@ -35,4 +35,26 @@ def test_sti_support_with_from_subclass
...
+      Object.const_set 'Subclass4', Class.new(model)
+
+      node1 = Subclass3.create!
+      node2 = Subclass4.create! :parent => node1
+      node3 = Subclass3.create! :parent => node1
+      node4 = Subclass4.create! :parent => node3
+      node5 = Subclass3.create! :parent => node4
+
+      assert_equal 2, node1.reload.children_count
+      assert_equal nil, node2.reload.children_count
```
kbrock, line 51: odd that this would be `nil` and not 0.
Do we need to put in a `to_i` in there or something?

Do you have a theory why it is happening?

test/concerns/sti_support_test.rb
```diff
@@ -35,4 +35,26 @@ def test_sti_support_with_from_subclass
       Object.send :remove_const, 'SubclassWithAncestry'
     end
   end
+
+  def test_sti_support_for_counter_cache
+    AncestryTestDatabase.with_model :counter_cache => true, :extra_columns => {:type => :string} do |model|
+      Object.const_set 'Subclass3', Class.new(model)
```
kbrock, line 41: Is there a reason why you did not use `Subclass1`?

If there were issues with reusing those class names, please let me know and I'll fix any issues.

kbrock (commented):
yay. this is green.

just a few questions.
Not saying that should block moving forward but they are a little concerning.

test/concerns/sti_support_test.rb
mattvague: Oh yes sorry I should have mentioned that, using the same class names was causing the error:

```
ActiveRecord::SubclassNotFound: Invalid single-table inheritance type: Subclass1 is not a subclass of Subclass1
```

test/concerns/sti_support_test.rb
mattvague: I _think_ it's because we didn't set a default value when setting up the children_count column on the `test_nodes` table https://github.com/stefankroes/ancestry/blob/1d0255133777f6710af32ae7b89a85aa62fa3971/test/environment.rb#L77

test/concerns/sti_support_test.rb
mattvague: My guess is that Rails is caching the constant name in the previous spec but it's not being wiped after... not sure how that part of Rails works though

kbrock:
ok. sorry. could you undo your last commit and squash your other commits? lets get this green and in.
I will circle back if I want to fix these constants. that should not be your concern.

Most important is to get this PR in and release another version of ancestry

mattvague:
>Could you undo your last commit and squash your other commits? Let's get this green and in.
I will circle back if I want to fix these constants. that should not be your concern.

Done!

mattvague:
🎈

PR #564, opened 2021-12-25
add rails 7 support

chenillen:


kbrock:
thank you.
I'm hoping to get rails 6.1 support before getting rails 7.0

We mostly have 6.0 but I'm pretty sure the `children.create!` is not working

petergoldstein:
@kbrock Is additional work still required to get Rails 6.1 working (it appears to be passing in CI).  Would you be open to a Rails 7 (and Ruby 3.1) PR at this point?

kbrock:
sorry, that this has not been merged yet.

I merged #577 which is similar/the same

kbrock:
There were strange warnings around 6.1. Specifically `children.create`, since the way rails works for `create` on top of a scope has changed a bit.

Please keep an eye on that in your project, but it looks like it is working.

Issue #273, opened 2016-03-18
NoMethodError: undefined method `ancestry'


djass:
When I try to save my object in database. Of course I migrated up my database.
I done that :

```
a = Component.new
a.save
```

And I have this error : NoMethodError: undefined method `ancestry' for #Component:0x00000005c95f68

```
My model is like this :
class Component < ActiveRecord::Base
has_ancestry
end
```

My migration is like this :

```
class AddAncestryToComponent < ActiveRecord::Migration
def change
add_column :components, :acst, :integer
add_index :components, :acst
end
def down

  remove_column :components, :acst
  remove_index :components, :acst

end
end
```

Please help me, what I done wrong ?

voke:
The column should be called `ancestry` and be of type `string` or `text`. But if you want to use a different name see the `ancestry_column` option.

kbrock:
thanks @voke I didn't see this question

kbrock:
Are you all set @djass ?

PR #565, opened 2021-12-28
Update README.md

benkoshy:
### What is this PR?

* Add detail re: customizing column names.

### Why is it required?

Because it may not be convenient to name the column: 'ancestry' - I imagine (perhaps incorrectly) that many would prefer to customize the column name to something better suited to their use-cases (e.g. https://github.com/stefankroes/ancestry/issues/273).

I hope this change is helpful - if it is not, feel free to close without compunction etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ $ bundle install
 
 ```bash
 $ rails g migration add_ancestry_to_[table] ancestry:string:index
+# or use different column name of your choosing. e.g. name:
+# rails g migration add_name_to_[people] name:string:index
 ```
 
 *   Migrate your database:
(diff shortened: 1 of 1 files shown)


kbrock:
Hello @benkoshy 

Thank you for the change. I will merge.

I've always been curious why someone would use a column other than `ancestry`

In theory, I understand the request. Especially when the column already exists in the database.
But I have yet to hear the reasoning from someone who actually needed the option.

Also, I fear that there is not a good enough test suite around that column so I wonder if it works in practice.

benkoshy:
> I've always been curious why someone would use a column other than ancestry

- it aids in conceptual understanding. e.g. `Car.where(ancestry: "123")` might be less preferable to `Car.where(model:  "Ford/T-model/1918/version_2")`

> Also, I fear that there is not a good enough test suite around that column so I wonder if it works in practice.

I had actually reverted to the `ancestry` name, after gaining more familiarity with the gem. by all means - feel free to revert if it is not beneficial.

I thank you for your work and efforts on maintaining this library.

kbrock:
well, it is good to document the api as it is. So thanks.

So I appreciate the PR and your followup response

PR #566, opened 2021-12-28
Use `ensure` to revert @disable_ancestry_callbacks

daniloisr:
Hello, I ran into a problem because I need the return of the block passed to `without_ancestry_callbacks`:

```rb
# always returns false instead of the result of inner_block
def some_method
  without_ancestry_callbacks { inner_block } 
end
```

So I think that `ensure` is a good fit here.

Note, this is probably a breaking change (although I think no one is relying on the result of `without_ancestry_callbacks` right now), so should we create a new method for this?

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -297,6 +297,7 @@ module Ancestry
     def without_ancestry_callbacks
       @disable_ancestry_callbacks = true
       yield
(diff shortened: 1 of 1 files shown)


kbrock (approved):
LGTM. just running the tests.

ping me if I do not merge this soon.

kbrock:
kicking

PR #571, opened 2022-01-06
Materialized path2

kbrock:
This introduces another serialization technique for the ancestry column

Will probably be converting these methods over to helper methods that do not call `ancestry_column`, so I'm not sure if introducing this helps or just introduces more gook to change

### DIFF
--- a/.github/workflows/run_test_suite.yml
+++ b/.github/workflows/run_test_suite.yml
@@ -70,7 +70,10 @@ jobs:
       - name: run sqlite tests
         env:
           DB: sqlite3
-        run: bundle exec rake
+        run: |
+          # for this database, run all the forms of our tests
+          bundle exec rake
+          STRATEGY=materialized_path2 bundle exec rake
 
       - name: run pg tests
         env:
@@ -88,6 +91,7 @@ jobs:
           # psql -c 'create database ancestry_test;' || echo 'db exists'
           bundle exec rake
 
+          STRATEGY=materialized_path2 bundle exec rake
       - name: run mysql tests
         env:
           DB: mysql2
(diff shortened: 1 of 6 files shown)


kbrock:
@d-m-u you have a comment on whether this is just adding complexity now or you think it moves us forward on allowing plug and play materialized path strategies?

kshnurov:
I guess this one wasn't tested? It's not even [loaded](https://github.com/stefankroes/ancestry/blob/master/lib/ancestry.rb#L6), you [can't inherit a module](https://github.com/stefankroes/ancestry/blob/master/lib/ancestry/materialized_path2.rb#L5) (`syntax error, unexpected '<' (SyntaxError)  module MaterializedPath2 < MaterializedPath`), the proper `ROOT = '/'` is not set in `materialized_path2`, and [some tests](https://github.com/stefankroes/ancestry/blob/master/test/concerns/validations_test.rb#L7) will inevitabely fail with `materialized_path2`.

kshnurov:
Fixed most (all?) of the issues in [this branch](https://github.com/kshnurov/ancestry/tree/materialized_path2_fix).
Failing tests should be updated or removed (like the ones that use `ancestry=`, I think this method should be private).

In my opinion, the new strategy is better because it uses 1 condition instead of 2.

kbrock:
Yea, I just found a few errors in here while I was refactoring something. Think it ends up with one too many slashes too

I had run a bunch of tests on this but looks like I botched it when moving it across

kshnurov:
Waiting for you to approve [this PR](https://github.com/stefankroes/ancestry/pull/589), and I'll send a fix for this one.

PR #577, opened 2022-06-09
Rebased #564 with Ruby 3.1 support and updated CI

petergoldstein:
This is a rebased version of #564 .

In addition to the rebase it includes:

1. Changes to the GitHub Actions file to run Rails 7 / Ruby 3.1 in the CI matrix
2. Changes to `test/environment.rb` to support `database.ci.yml` file loading in Psych 4 (by explicitly enabling aliases)
3. Removing the now unused .travis.yml, since there's no reason to keep that in sync.

I'm not sure why the VERSION value was bumped in the original PR, but I've left it as it was in the original PR.


petergoldstein:
This runs green on my fork.

kbrock:
Just to be clear, we are expecting that rails 7.0 may fail.
But we want the failure to be there so we can fix it?

PR #585, opened 2022-07-10
Fix non-default primary_key in find_by and where queries

Zhong-z:
similar to #575 but trying to fix all of them


kbrock:
thanks @Zhong-z

kbrock:
so confused why I did not merge this ages ago

PR #588, opened 2022-09-01
add `update_strategy` documentation

victorfgs:
It was missing from the original doc and it made me look in the codebase to see how it works and how to use it. It seems to me to be a nice addition to the official doc.

Thanks for the great gem!


kbrock:
thanks @victorfgs

PR #589, opened 2022-09-05
update_descendants_with_new_ancestry in after_update

kshnurov:
This simple code will not work as expected when you update a parent's slug - descendants won't have the slug changed.
```
class Model
  before_save :update_slug

  def update_slug
    self.slug = [parent.slug, self.slug].join('/')
  end
end
```

That is because `update_descendants_with_new_ancestry` is run with `before_save` and the `parent` method loads an unchanged version from the database.
This PR moves `update_descendants_with_new_ancestry` to `after_update`.

There's no tests for such case right now, but I've tested it with my code and it works. Testing with `update_strategy = :sql` would be appreciated, though the change is pretty simple and should just work.

### DIFF
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -45,8 +45,8 @@ module Ancestry
       # Validate that the ancestor ids don't include own id
       validate :ancestry_exclude_self
 
-      # Update descendants with new ancestry before save
-      before_save :update_descendants_with_new_ancestry
+      # Update descendants with new ancestry after update
+      after_update :update_descendants_with_new_ancestry
 
       # Apply orphan strategy before destroy
       before_destroy :apply_orphan_strategy
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -5,15 +5,15 @@ module Ancestry
       errors.add(:base, I18n.t("ancestry.exclude_self", class_name: self.class.name.humanize)) if ancestor_ids.include? self.id
     end
 
-    # Update descendants with new ancestry (before save)
+    # Update descendants with new ancestry (after update)
     def update_descendants_with_new_ancestry
       # If enabled and node is existing and ancestry was updated and the new ancestry is sane ...
       if !ancestry_callbacks_disabled? && !new_record? && ancestry_changed? && sane_ancestor_ids?
         # ... for each descendant ...
-        unscoped_descendants.each do |descendant|
+        unscoped_descendants_before_save.each do |descendant|
           # ... replace old ancestry with new ancestry
           descendant.without_ancestry_callbacks do
-            new_ancestor_ids = path_ids + (descendant.ancestor_ids - path_ids_in_database)
+            new_ancestor_ids = path_ids + (descendant.ancestor_ids - path_ids_before_last_save)
             descendant.update_attribute(:ancestor_ids, new_ancestor_ids)
           end
         end
@@ -133,8 +133,8 @@ module Ancestry
       ancestor_ids + [id]
     end
 
-    def path_ids_in_database
-      ancestor_ids_in_database + [id]
+    def path_ids_before_last_save
+      ancestor_ids_before_last_save + [id]
     end
 
     def path depth_options = {}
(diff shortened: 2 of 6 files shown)


kshnurov:
There's a new `descendants_by_ancestry(ancestry)` scope, should we add it to the docs?

kshnurov:
@kbrock have you had a chance to look into this?

kbrock:
@kshnurov thanks for the ping.

When I modify values in a model like a field (`slug`), I tend to do those in before validation. So regardless of the `before_*` orders, the record is complete.

If you move the code into a `before_validates`, does that fix it.

Also, is it possible that `update_slug` brings in `name` or `tag` rather than `slug`:

```ruby
  def update_slug
    self.slug = [parent.slug, self.name].join('/')
  end
```

kshnurov:
Hi @kbrock 

> When I modify values in a model like a field (`slug`), I tend to do those in before validation. So regardless of the `before_*` orders, the record is complete.

Me too, this is just an example, doesn't matter which callback you use, I use `friendly_id` gem, which uses both.
The issue here is the parent object.
 
> Also, is it possible that `update_slug` brings in `name` or `tag` rather than `slug`:

This is just a mockup example, a typical use would be some kind of title or name, but once again - it's not the issue here.
The issue is `parent.slug` being incorrect, since the change in parent is not saved yet, and we load an unchanged version here.

kbrock:
Hello @kshnurov 

Thank you for the clarification. it sounds like we are on the same page.

There is definitely an issue when there are multiple version of `parent` loaded into memory. a [test](https://github.com/kbrock/ancestry/blob/master/test/concerns/hooks_test.rb#L5) was created for similar reasons. This update case is a nice addition.

I'll take another look at this. This does look like it is just a conversion from one hook to another.

> > I need to document a good way to build a field that is composed of the ancestry name values rather than id values.

For this case, introducing a database only update tends to works better, but that will not update in memory nodes, so that has the potential to introduce edge cases.

/ASIDE: Wonder if this pattern is common enough to warrant introduction into the gem itself.

kshnurov:
> I'll take another look at this. This does look like it is just a conversion from one hook to another.

Yes, since we should update descendants **after** the record is saved, not before.
I've added a [simple test](https://github.com/stefankroes/ancestry/pull/589/files#diff-3c1d68efc6790af74a4e2de13628a7d617e930b4d3011aee234355d7157f7b2c), it fails on the main branch and passes with this PR.

Also, there's a reason why I used `before_save`: `before_validation` is not triggered by `update_attribute`, which is [used to update descendants](https://github.com/stefankroes/ancestry/blob/master/lib/ancestry/instance_methods.rb#L17). Probably should add a note to the docs about that.

> For this case, introducing a database only update tends to works better, but that will not update in memory nodes, so that has the potential to introduce edge cases.

Database only update obviously won't work with this case or any other callbacks, like updating data in your search DB, and you'll end up with inconsistent data. I think they're evil and shouldn't exist, because they will break pretty much everything, including a simple cache (since `updated_at` isn't changed). I don't think saving a few DB transactions when changing ancestry (which sounds like a rare case) is worth the risk and extra caution required, so I'd remove them completely.

> /ASIDE: Wonder if this pattern is common enough to warrant introduction into the gem itself.

Using parent values is inevitable if you have friendly urls like `/a/b/c/d`, which is quite common.

kshnurov:
@kbrock as you can see tests fail with PG, since DB-only updates don't trigger callbacks. Skipped that test for sql strategy.

kshnurov:
@kbrock is there anything holding you from merging this?

kbrock:
@kshnurov I appreciate where you are coming from.

I'm concerned that after update will need to save the child records twice instead of once.

Is your thought that we will very rarely update the ancestry, so having that operation a double save is not bad?

Or I guess it is possible that the primary record will be changed, maybe double saved, then the followup records will only get only updated once?

kshnurov:
@kbrock why will it save twice? It won't, child records will be updated only once - when parent `ancestry` is changed. Saving/touching parent without changing ancestry won't trigger anything.

If someone adds their own callback to update children on parent change, it's up to them to avoid running callbacks twice in such rare case. Anyway there won't be any double-saving to DB since all changes will happen on the first run, second run won't change the record.

kshnurov:
@kbrock 1,5 months and this simple PR didn't move. Should we consider this gem unmaintained?

kshnurov:
@kbrock ?

kbrock:
Sorry for the delay.

Updating the callback order is potentially a breaking change for the hundred(s?) of projects that depend upon this gem.

I am looking into fixing this for sql based updates

PR #597, opened 2023-01-05
Fix materialized_path2

kshnurov:
materialized_path2 was completely broken [from the beginning](https://github.com/stefankroes/ancestry/pull/571#issuecomment-1245091316).
This PR fixes it and renames `strategy` to `ancestry_format` since it's not really a strategy, it doesn't change any behavior.

### DIFF
--- a/.github/workflows/run_test_suite.yml
+++ b/.github/workflows/run_test_suite.yml
@@ -78,7 +78,7 @@ jobs:
         run: |
           # for this database, run all the forms of our tests
           bundle exec rake
-          STRATEGY=materialized_path2 bundle exec rake
+          FORMAT=materialized_path2 bundle exec rake
 
       - name: run pg tests
         env:
@@ -95,8 +95,8 @@ jobs:
           # db container is currently creating the database
           # psql -c 'create database ancestry_test;' || echo 'db exists'
           bundle exec rake
+          FORMAT=materialized_path2 bundle exec rake
 
-          STRATEGY=materialized_path2 bundle exec rake
       - name: run mysql tests
         env:
           DB: mysql2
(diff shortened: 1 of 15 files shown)


kbrock (commented):
This PR is great.

I do have the goal of going away from all these `cattr_accessor` values. To be honest, this sounds minor but it has been my main sticking point. It is the reason I am dragging my feet.

Rails has helper methods that mixed into the class that take parameters. And then dynamically creates methods with the special variable (e.g.: delimiter, ancestry_column) passed in as a parameter.

We, on the other hand, mix in a class that access the special variable via cattr.

If we go towards this other behavior, then we can have multiple ancestry columns per record and I believe it would actually simplify our code.

Is there a way to change this code so we are not introducing yet more cattrs?

kshnurov:
@kbrock these parameters were previously hard-coded, which is definitely not good. Now they are used both in other modules (like the `materialized_path2` or `materialized_path_pg`) and application code (like the provided migration).

It probably makes sense to change them to cattr_reader and make some of them private, but other than that it's just defining some variables instead of constants. What's the other option?

kshnurov:
@kbrock here's a take on `cattr_accessor`: https://github.com/kshnurov/ancestry/commit/493f4eefe1a3e197c35c21c4023394664295e1f9
It has breaking changes: all writers are removed (you shouldn't change ancestry options on-the-fly, that'll lead to unexpected behavior) and all instance accessors are removed.

kshnurov:
@kbrock how about releasing 4.3 on RubyGems? :)

kshnurov:
@kbrock 4.2 is still the latest version on RubyGems, do you have access to release a new version?

kbrock:
@kshnurov Yes, I have access.

- I wanted to fix your issues with #601 - It started as ascii varchar, and prefer to get it as close to the original field as possible. But I would like to document the `binary` case that you mention. I would like that case better if the sql generated were text and not escaped binary.
- I want to double check if #589 introduced something bad
- I need to find the time to update the changelog
- add a postgres sql only for update_descendants_after_update

Issue #603, opened 2023-02-24
release next version


kbrock:
- [x] https://github.com/stefankroes/ancestry/pull/607
- [x] double check if #589 introduced something bad
- [x] update the changelog
- [x] Enable test for [hooks_test.rb:35](https://github.com/stefankroes/ancestry/blob/721699e05244619cf7a585ed9f484a92dba1c297/test/concerns/hooks_test.rb#L35) -- punt. will do later

kbrock:
probably not for this release
want to setup config for 2 things:

- [ ] materializedpath2 is the default
- [ ] no self returned from siblings
- [ ] warning message that you need to set these configs for upgrades via `gemspec.post_install_message`

kshnurov:
Well, the main branch is already at 4.3 with a proper changelog. All you need it to update RubyGems.

#601 is just a README change, nothing to release here.

materializedpath2 is the default - I agree, but it should be a major version bump.

kshnurov:
> It started as ascii varchar, and prefer to get it as close to the original field as possible. But I would like to document the binary case that you mention. I would like that case better if the sql generated were text and not escaped binary. Want to change mysql to a binary column. (character set=ascii would be preferable but just too hard)

I guess you should just [read](https://dev.mysql.com/doc/refman/8.0/en/charset-general.html) about database collations/character sets, what they do and when they are needed. This gem doesn't need ascii, `C`, `latin` or whatever collation there is. `binary` is all that's needed, it's faster and takes less space.

kshnurov:
> no self returned from siblings

Why? It would be a very simple, but breaking change.
Whoever needs that, should just do `self.siblings.without(self)`, that'll produce a proper SQL.

kshnurov:
> track down test disabled for postgres extension. Thought it was for update_descendants_after_update. I possibly already fixed this. can't find it

It's [this one](https://github.com/kshnurov/ancestry/blob/721699e05244619cf7a585ed9f484a92dba1c297/test/concerns/hooks_test.rb#L35). I'm strictly against this SQL strategy anyway, there's nothing good in skipping all Rails callbacks. It's not even updating the `updated_at` - cache won't be invalidated.
Btw, have no idea why the README says it's PG only, should also work with MySQL.

kbrock:
re #601: In postgres, `binary` seems to use indexes better than the collation mode. So this is an improvement. But the sql generated is hard to read:

```sql
[...]

PR #607, opened 2023-03-01
Improve documentation around ancestry field and collation

kbrock:
postgres: binary data type is not currently working
mysql: binary data type or string with binary collation is working

Had wanted to go with binary for all, but test would not pass

this updates the comments introduced in #601

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,9 @@ NOTE:
 
 # Installation
 
-Follow these simple steps to apply Ancestry to any ActiveRecord model:
+Follow these steps to apply Ancestry to any ActiveRecord model:
 
-## Install
-
-* Add to Gemfile:
+## Add to Gemfile
 
 ```ruby
 # Gemfile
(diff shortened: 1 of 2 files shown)


kbrock:
@kshnurov I updated the gem to use `binary` for postgres. The test suite puked.

Staying with the original comments, but put in more documentation to hopefully get this running correctly.

kshnurov:
> @kshnurov I updated the gem to use binary for postgres. The test suite puked.

Ok, after some digging it turns out `pg` gem doesn't unescape `bytea` by default and postgres returns it in hex format.
Here's a solution for postgres, it'll should also solve [this](https://github.com/stefankroes/ancestry/issues/603#issuecomment-1448800893)
```
ALTER DATABASE dbname SET bytea_output to 'escape';
```

With this one, tests are green unless we use `update_strategy = :sql`.
`regexp_replace` isn't possible with binary columns. I'd suggest removing `:sql` update strategy completely.

kbrock:
@kshnurov For a binary ancestry column, what else do I need to specify to set a length? And what value should I use?

```
 Mysql2::Error: BLOB/TEXT column 'ancestry' used in key specification without a key length (ActiveRecord::StatementInvalid)
```

kshnurov:
> For a binary ancestry column, what else do I need to specify to set a length? And what value should I use?

IDK if you missed [this comment](https://github.com/stefankroes/ancestry/pull/601#issuecomment-1437888440)
It definitely shouldn't be a BLOB.

kbrock:
update:
- provided simplified column information and a section with probably way too much detail. After time, this can probably be simplified.
- Added details about using/preferring materialized_path2
- Added configuration file step to default to materialized_path2

PR #612, opened 2023-03-02
Add default ancestry format and primary key format

kbrock:
These tend to be the same across a whole database.

The primary key format is used for databases with uuids for the primary key The ancestry format makes it easy for a new user to get up and running using a different ancestry encoding format

### DIFF
--- a/lib/ancestry.rb
+++ b/lib/ancestry.rb
@@ -12,6 +12,8 @@ I18n.load_path += Dir[File.join(File.expand_path(File.dirname(__FILE__)),
 
 module Ancestry
   @@default_update_strategy = :ruby
+  @@default_ancestry_format = :materialized_path2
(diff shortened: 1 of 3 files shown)


kbrock:
@kshnurov do you know how to use `mattr_accessor` here?

I feel this is probably the easiest way to get users up and running using the new materialized path2.

kbrock:
if you have an answer, we can circle back later. I want to get this merged, so I can use it in documenting ancestry collation and get us closer to a gem release

kshnurov:
You're instantly merging things left and right without even thinking and reviewing. What the hell is going on?

kbrock:
I wanted to simplify our readme and this made it easier.
Yes, I did find a bug after the fact. I am putting in a PR to fix.

PR #613, opened 2023-03-03
Partial tree sort tests

kbrock:
extracted from https://github.com/stefankroes/ancestry/pull/415

add documentation to tests.

Unsure if we can assume that nodes passed into `sort_by_ancestry` are in a sane order.

### DIFF
--- a/test/concerns/sort_by_ancestry_test.rb
+++ b/test/concerns/sort_by_ancestry_test.rb
@@ -1,12 +1,11 @@
 require_relative '../environment'
 
 class SortByAncestryTest < ActiveSupport::TestCase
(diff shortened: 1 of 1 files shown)


kbrock:
update:

I had thought that pagination would ensure that you don't end up with cousins and not at least one aunt/uncle. It looks like even when using `order_by_ancestry_and(:rank)` that you can definitely get cousins without any parents.

Dropped that set of test changes.

example for reference:

- 1
  - 6
- 2
  - 8 
- 3
- 4
- 5
  - 7

`order(ancestry,id)`: 1, 2, 3, 4, 5, 6, 8, 7

since 6,8 are so far from 1,2, pagination can definitely throw them seperate.

PR #600, opened 2023-02-15
Refactor options storage

kshnurov:
All writers are removed (you shouldn't change ancestry options on-the-fly, that'll lead to unexpected behavior) and all instance accessors are removed.
We do not mention `ancestry_column` or `ancestry_delimiter` accessors in the docs, but if someone is using them this will be a breaking change for them.
Since this is an internal API, I'm not sure if we should bump a major version. What do you think?

PR #617, opened 2023-03-04
Drop Orphan strategy class attribute

kbrock:
This is part of a push to remove so much global state stored in the class.
Some parts pulled from #600 

`orphan_strategy` getters and setters on the class were never part of the public API but necessary for an internal implementation. Removing this internal mechanism that has potentially leaked to external implementations.

Currently the supported usage pattern for `orphan_strategy` is still:

```ruby
class Model < ActiveRecord::Base
  has_ancestry orphan_strategy: :destroy
end
```

Dropping accessing and changing the value directly in the class:

```ruby
class Model < ActiveRecord::Base
  has_ancestry
  # removing: self.orphan_strategy = :destroy
  # removing: puts "changed to #{orphan_strategy}"
end
```

---

I did go through github and no projects using ancestry referenced `orphan_strategy` or `orphan_strategy=`.

A number of the projects did call `has_ancestry` with `:orphan_strategy`, as well as `acts_as_tree`. Which is good. that means this feature is in use and those projects only use the public api.

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -21,17 +21,6 @@ module Ancestry
       end
     end
 
-    # Orphan strategy writer
-    def orphan_strategy= orphan_strategy
-      # Check value of orphan strategy, only rootify, adopt, restrict or destroy is allowed
-      if [:rootify, :adopt, :restrict, :destroy].include? orphan_strategy
-        class_variable_set :@@orphan_strategy, orphan_strategy
-      else
-        raise Ancestry::AncestryException.new(I18n.t("ancestry.invalid_orphan_strategy"))
-      end
-    end
-
-
     # these methods arrange an entire subtree into nested hashes for easy navigation after database retrieval
     # the arrange method also works on a scoped class
     # the arrange method takes ActiveRecord find options
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -14,9 +14,6 @@ module Ancestry
       end
 
       orphan_strategy = options[:orphan_strategy] || :destroy
-      if ![:rootify, :adopt, :restrict, :destroy].include?(orphan_strategy)
-        raise Ancestry::AncestryException.new(I18n.t("ancestry.invalid_orphan_strategy"))
-      end
 
       # Create ancestry column accessor and set to option or default
       cattr_accessor :ancestry_column
@@ -58,10 +55,6 @@ module Ancestry
       update_strategy = options[:update_strategy] || Ancestry.default_update_strategy
       include Ancestry::MaterializedPathPg if update_strategy == :sql
 
-      # set orphan strategy
-      cattr_reader :orphan_strategy
-      class_variable_set :@@orphan_strategy, orphan_strategy
-
       # Validate that the ancestor ids don't include own id
       validate :ancestry_exclude_self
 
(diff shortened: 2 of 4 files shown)


kshnurov:
@stefankroes pay attention to this and please stop breaking this gem.

@kbrock what you've done here is a breaking change and you didn't even mention it in the changelog.
First of all, people may have been skipping or overriding `apply_orphan_strategy` callback, but you've renamed it without any second thought.
Also, they could've been using `orphan_strategy` accessor in other places, but you've just wiped it out.

PLEASE STOP FUCKING UP THIS GEM.

kbrock:
@kbrock what you've done here is a breaking change and you didn't even mention it in the changelog.

I will be sure to include this change when I do write up the changelog.

> First of all, people may have been skipping or overriding `apply_orphan_strategy` callback, but you've renamed it without any second thought.

Thank you for pointing this out. 

Do you override this method?

If someone didn't want to use our strategy, they can just pass `orphan_strategy: none` and define their own `before_destroy` hook. Since this method is not part of the public interface, I didn't realize they would override it. I'm not totally sure what a public interface is now a days since ruby lets you do anything you want. I will look into fixing this.

I see examples of that method being overridden in github:

- [taxon.rb](https://github.com/inaturalist/inaturalist/blob/48ca16d09d729fd897aaca7054ff363fddf606e6/app/models/taxon.rb#L1898) and other repos seem to have copy/pasted code this code.
- [project.rb](https://github.com/glittergallery/GlitterGallery/blob/fded5eef7548e6845e604b7e13175a681bbf115c/app/models/project.rb#L44)

> Also, they could've been using `orphan_strategy` accessor in other places, but you've just wiped it out.
>

You can't have it both ways. This was inspired by #600 - that PR drops the `orphan_strategy` setter and getter. As I stated, I found no code that used this. While that is by no means a thorough investigation of all code that exists, it is a reasonable test. Since the developers are setting this, I didn't see need to tell them what they already know.

> PLEASE STOP FUCKING UP THIS GEM.

Really? Do you feel this is the best way to work with another individual? When you contribute to other github projects, do you act this way?

Your tone is so unprofessional, that I am having trouble taking you seriously.

kshnurov:
> Your tone is so unprofessional, that I am having trouble taking you seriously.

I obviously don't want to work with you, because that means fixing your bad code or stopping you from merging it, which I've already done half a dozen times. Your work on this gem in the past month is absolutely unprofessional, junior level code.
You're consistently doing things a good programmer or maintainer would never do.
I'm done defending this gem, it's such a waste of time.

stefankroes:
@kshnurov You keep dragging me into this, but you should know that for all intents and purposes, I consider @kbrock the owner of this gem. He has been maintaining it for **7 years** without any help from me and without him it would have been dead years ago ❤️ 

As for me, I think @kbrock has been extremely patient and constructive at every turn. Regardless of what happened and who's right from a technical point of view, your aggressive and negative comments are not wanted here. They should end right here and now.

However, your help and expertise are wanted here very much. I suggest we close this thread. I will delete and/or report any further aggressive or negative messages. You are very much welcome to open new issues or PRs in a positive and constructive manner.

kshnurov:
> I consider @kbrock the owner of this gem
> Regardless of what happened and who's right from a technical point of view, your aggressive and negative comments are not wanted here

@stefankroes you're the owner of this repo, either take your responsibility for it or change the owner, if you don't want to be taken responsible. Avoiding responsibility and hiding from the negative truth is not what adults do.

kbrock:
There is no `orphan_strategy: none` option and it would not make sense since doing nothing would make ancestry for all the children invalid.

The best way to change the behavior is to extend this hook.

I have already put together a fix but am researching the usage patterns in github to ensure this test reflects the user's needs.

I will also add this method to the public interface / documentation.

Issue #586, opened 2022-07-12
`descendants_count` counter cache?


ptorrsmith:
We're wanting to implement a good way to cache the `descendants_count` on nodes in a similar way that `children_count` is set and maintained.

Has anyone attempted extending Ancestry to do this?

Currently we're looking at a pragmatic approach of having an async background job get queued up to update the ancestors of any added/removed node, but thinking this seems like it would be a nice extension for Ancestry.

kbrock:
The counters are added as a one off and hardcoded into the code itself.
They end up needing to be hardcoded into the postgres only extension as well.

It would be great to be able to mixin these changes for those because both types of counters are extensions for the most part.

If you notice, they are defined in the postgres code and has_ancestry code.

sorry, thought I posted this a while ago.

kbrock:
status update:

Making ok progress at updating the api so we can update the descendants count (hopefully for ruby and sql strategies. Thought I think I'd keep with ruby strategy for now)

kbrock:
@ptorrsmith curious. Are you still using ancestry? Did you ever end up implementing this? Did you go the background job approach?

ptorrsmith:
> @ptorrsmith curious. Are you still using ancestry? Did you ever end up implementing this? Did you go the background job approach?

Hi. Yes still following this.  We ended up baking a post-run call to recalculate it and cache it.  Would be good to have this as part of ancestry.

kbrock:
@ptorrsmith finally got this out. Have you been using github branch up until now? Did you use the sql focused update code in there?

The sql does a +1 / -1. Maybe having it do a pure calculation may be better.

PR #307, opened 2016-11-04
Counter cache

hw676018683:
Introduce child_counter_cache
It counts the number of direct children

close https://github.com/stefankroes/ancestry/issues/303

PR #618, opened 2023-03-04
fix increase_parent_counter_cache callback

### LINKED ISSUE #584: find and fix corrupted children counts
We have a rather large database, and one of our clients wrote in about a bug that we determined to be a result of children_count being wrong for many many users ( we aren't really sure how many) Is there something I can do to regenerate all children counts in our database?

kbrock:
no longer incrementing counter when callbacks are disabled (to make increase and decrease consistent)

Update parent_counter_cache now respects @_trigger_update_callback

Not sure if update follows the same rules as destroy introduced in 63b42f53 But would like both trees to follow the same logic

There are reports of incorrect cache counts. So when insert and destroy have different guard clauses, it is suspect.

I admit that I am not totally familiar to the core use case so I will need to more into why this was introduced in #307 - trying to resolve #584 (and may be able to provide advice for #586 )

TODO:

- introduce tests that corrupt the counter caches see #584 (and mildly related: #586 )
- introduce test with rollback
- introduce STI parents for test (parent should be parent class and child node should be a child class of the parent)


kbrock:
Feature (not fixing the bug):
- see if we can use `REPLACE` instead of `REGEX_REPLACE`. regex buys us anchoring at the beginning of the line (which is needed if we want to support materialized_path), but `REPLACE` buys us binary support.
-

PR #619, opened 2023-03-04
Counter cache column

kbrock:
code was pulled out of #600

- Drop unnecessary `_counter_cache_column`, and remove all the `to_s` that were used.

May merge with #618 - but at this time, refactoring the code and fixing the counters that get askew seem like different problems

### DIFF
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -87,12 +87,7 @@ module Ancestry
       # Create counter cache column accessor and set to option or default
       if options[:counter_cache]
         cattr_accessor :counter_cache_column
-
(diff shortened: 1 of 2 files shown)


kbrock:
@kshnurov A few questions:

- was there a particular use case you were protecting by adding the `parent_id` checks?
- It looks like those checks are already made in the callback definitions.

I did notice that `ancestry_callbacks_disabled?` was not checked in the `increase_parent_counter_cache`. Not sure if these are the root of #618. Will probably move from this PR and put into that one since it is more of a potential bug fix rather than a refactor.

kbrock:
update:
- dropping parent_id checks from callbacks. leaving that code in #618 so we can apply if it is necessary

Issue #87, opened 2012-02-28
using has_ancestry in subclass throws error


jtromp:
When I add "has_ancestry" in a subclass of another class (that inherites from ActiveRecord:Base), an error is thrown in development mode:

Mysql::Error: Table 'AUTHORABLE.groups' doesn't exist: SHOW FIELDS FROM `groups`

The class hierarchy in this case is: ActiveRecord::Base > Authorable > Group

The Group class has "has_ancestry" defined. Is this expected behaviour, or is there a solution? I could not find anything related to this in the documentation.

kbrock:
Hello @jtromp 

Not sure if you have github alerts turned on, and you probably moved on from this application a long time ago.

But I am hoping you might remember a little bit about this application from 10+ years ago.

Do you remember if you liked how this model turned out? Can you guess if putting `has_ancestry` at the top level would have worked out just as well?

I'm trying to understand if this is an important feature to keep, and thought I would make a long shot in reaching out to you.

Thanks,
Keenan

PR #620, opened 2023-03-05
Drop ancestry_base_class

kbrock:
active_record defines base_class as the top of an STI tree.

ancestry had a similar concept. It was the class where ancestry was defined.
Since ancestry originally called it a base_class,
there was a conflict and in 4e57f32e1758 it was renamed to ancestry_base_class
 
PRs of note:
- #86
- #87

The ancestry_base_class is used quite often to make sure the proper class
is accessed to lookup values.

I'm not sure the use case for defining ancestry in the middle of a hierarchy,
but it doesn't seem to work in a reasonable manner. (UPDATE: the tests were faulty. I'm fixing the tests so it is properly tested)

```
class  Class1 < ActiveRecord:Base
end
class Class2 < Class1
  has_ancestry counter_cache: :child_count
end

root = Class2.create!
node = root.children.create!

root.children == []
node.root == root
root.child_count == 0
```

I'm dropping this custom concept and using the standard rails concept
instead, as I think it should have been done originally.

---

Contemplating raising an error if a developer adds `has_ancestry` to a non `base_class`.

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -2,7 +2,7 @@ module Ancestry
   module ClassMethods
     # Fetch tree node if necessary
     def to_node object
-      if object.is_a?(self.ancestry_base_class)
+      if object.is_a?(self.base_class)
         object
       else
         unscoped_where { |scope| scope.find(object.try(primary_key) || object) }
@@ -11,7 +11,7 @@ module Ancestry
 
     # Scope on relative depth options
     def scope_depth depth_options, depth
-      depth_options.inject(self.ancestry_base_class) do |scope, option|
+      depth_options.inject(self.base_class) do |scope, option|
         scope_name, relative_depth = option
         if [:before_depth, :to_depth, :at_depth, :from_depth, :after_depth].include? scope_name
           scope.send scope_name, depth + relative_depth
@@ -40,9 +40,9 @@ module Ancestry
     # Get all nodes and sort them into an empty hash
     def arrange options = {}
       if (order = options.delete(:order))
-        arrange_nodes self.ancestry_base_class.order(order).where(options)
+        arrange_nodes self.base_class.order(order).where(options)
       else
-        arrange_nodes self.ancestry_base_class.where(options)
+        arrange_nodes self.base_class.where(options)
       end
     end
 
@@ -175,7 +175,7 @@ module Ancestry
     def restore_ancestry_integrity!
       parent_ids = {}
       # Wrap the whole thing in a transaction ...
-      self.ancestry_base_class.transaction do
+      self.base_class.transaction do
         unscoped_where do |scope|
           # For each node ...
           scope.find_each do |node|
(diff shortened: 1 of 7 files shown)


kbrock:
/cc @Fryguy could you take a peek.

With these changes, manageiq-core along with relationships work fine.

Can you think of negative implications of changing this?
Also, can you think of a reason why someone may even want to define `has_ancestry` in the middle? We do not do this.

Fryguy:
I can't really think of a valid reason.  The only thing that comes to mind, and it feels way outside the norm, is if you want different has_ancestry settings for different subclasses from the same base class, but then I'd think you'd need multiple separate ancestry columns as well.

kbrock:
@kshnurov do you have an opinion on this?

kbrock:
will introduce some of the other #600 changes and then revisit this.

I'm on the fence but leaning towards yes.

Fryguy:
This is a breaking change, right, so major version bump? (Or should we deprecate?)

kshnurov:
> @kshnurov do you have an opinion on this?

You're introducing a breaking change that adds/saves/solves nothing, but will break [someone's use cases](https://github.com/stefankroes/ancestry/issues/87).
What's the purpose, just to break something?

kbrock:
> > @kshnurov do you have an opinion on this?
> 
> You're introducing a breaking change that adds/saves/solves nothing, but will break [someone's use cases](https://github.com/stefankroes/ancestry/issues/87).

A simple, "Lets keep this. there is an issue that states it is needed" would suffice.

The phrase "I'm on the fence but leaning towards yes" means that I am unsure. That is why I reached out.

> What's the purpose, just to break something?

Don't be rude just to be rude. This adds nothing.
This is not the way to encourage me to ask for your input in the future.

kshnurov:
> A simple, "Lets keep this. there is an issue that states it is needed" would suffice.

You mentioned that issue in the PR description! I didn't expect you didn't even read it 🤦🏻‍♂️

> Don't be rude just to be rude. This adds nothing.
This is not the way to encourage me to ask for your input in the future.

You've thrown out my opinion, proposals and work a few times already without any objective reasoning. I consider that being rude, so the attitude is the result of that and a few new bad PRs that make this gem only worse.

kbrock:
I'll revisit this later

kbrock:
For my own reference

- `base_class` was around in rails 1.1.6 (2006)
- `base_class` was introduced in ancestry 1.1.2 (2009) in e3ea26a54e1f
  - This said to add STI support (not specific whether it is adding STI in the middle)
  - confused why `base_class` was overridden instead of leveraging existing column
  - was this name intentionally picked, or was it a coincidence
- `s/base_class/ancestry_base_class/g` in #86 
  - fixes conflicts with other gems
  - reason for replacing rather than just dropping the assignment of base_class?
  - fixes 87 as well. was this intentional?
- test to ensure mid STI support introduced in 4e57f32 fixed 87 : 0cc1c0f0
  - first reference in the code base that sti in the middle is a thing.

I guess I am stuck trying to figure out why someone would define an STI tree and only want to use part of it for an ancestry tree. Maybe there is a default scope in a portion of it?

Also stuck having a concept that seems to overlap with rails so closely. (especially for a feature that I do not think is valid. would prefer to not have extra concepts in the code)
---

I am only ever going to define this at the root level. So whether there is this variable does not affect me.

And if that premise is followed, it is not in the way for having multiple ancestry columns.

Issue #622, opened 2023-03-10
Falseful changelog, incorrect README and broken code in 4.3.0


kshnurov:
@stefankroes what is happening to this gem? Are you ok with this?
I can take over the repo and fix the mess created by @kbrock if you let me.

4.3.0 changelog states:
`* arrange is 3x faster and uses 20-30x less memory [#415](https://github.com/stefankroes/ancestry/pull/415)`

[The PR itself](https://github.com/stefankroes/ancestry/pull/415) is called `Make arrange_nodes more memory efficient`.
But `arrange` and `arrange_nodes` were not even touched. That's a straight lie!

README now contains [absolutely incorrect setup instructions](https://github.com/stefankroes/ancestry/blob/master/README.md?plain=1#L356-L391) that are either broken or will result in bad performance.
I've provided the correct ones [here](https://github.com/stefankroes/ancestry/pull/607) and [here](https://github.com/stefankroes/ancestry/pull/601), but @kbrock just doesn't want to listen and learn about database collations.

Also, both of the last 2 versions were released with broken code that just doesn't work as described.
[One I've fixed](https://github.com/stefankroes/ancestry/pull/597) and [the second one is still there](https://github.com/stefankroes/ancestry/pull/588).

kbrock:
@kshnurov your communication style is quite combative. These attacks are getting tiring. I'd rather be spending my time merging your PRs than fighting you.

You had the option to put in a PR or to attack me and call me a liar. You chose the latter.
It feels like you often choose the attacking route.

I could have merged a PR that you submitted without a second though, but here I am defending myself against your most recent attack. I'm sorry @stefankroes that you are dragged into this.

---

Thank you for finding another typo. We were working on `flatten_arranged_nodes` (and the alternative to `arrange_as_array`. As you'll notice, `arrange` is the common word there. Yes, I made a mistake. It seems like a reasonable mistake but I will change it to `sort_by_ancestry` to make sure no one is misguided. Saying I was intentionally lying is a bit of a stretch.

I'd imagine your frustration would be heightened since you have clearly stated that you want to delete this method from the code base. 

But we can't delete this method. It is needed. The original form has been in the wiki for a long time with people updating it. And just last week someone submitted a few issues wanting to do this very thing. I need to go in there and `s/arrange_as_array/sort_by_ancestry/g`. But again, you are capable of contributing instead of attacking.

[...]

PR #624, opened 2023-03-10
Tests more closely follow suggested configurations

kbrock:
## Test both `materialized_path`

~~I hadn't noticed that #597 dropped all the `materialized_path` testing.
Since a majority of the users are using `materialized_path`, we need to make sure this is working front and foremost~~

Setting the column to `nulls: true` only for `materialized_path`

### Before

We were testing postgres only with `:sql` update mode.

### After

This behavior has not changed. But now it is more explicit.

It seems fine to only test postgres with `:sql` update mode since
both mysql and sqlite paths are testing the traditional `:ruby` update mode.

## Test mysql with binary columns

Running mysql with both ascii and binary columns since we support both.

## Support matrix

- https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html
- https://www.ruby-lang.org/en/downloads/branches/
- https://guides.rubyonrails.org/maintenance_policy.html

Rails 5.2 and 6.0 (and the corresponding ruby 2.5 to 2.7) are basically EOL. Lets keep them until they hold us back.

Due to the way I've setup the matrix, I need to choose a single rails version to test with 7.0. I chose 2.2 since that is rail's suggested version.

This leaves us not testing ruby 2.3. If anyone knows how to add both 2.2 and 2.3, please share.

### DIFF
--- a/.github/workflows/run_test_suite.yml
+++ b/.github/workflows/run_test_suite.yml
@@ -37,19 +37,26 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # this matrix is driven by these sources:
+        # - https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html
+        # - https://www.ruby-lang.org/en/downloads/branches/
+        # - https://guides.rubyonrails.org/maintenance_policy.html
+        format: [materialized_path, materialized_path2]
+        activerecord: [70]
+        ruby: [3.1, 3.2]
+        # additional tests
         include:
+          # EOL 6/2022
           - ruby: 2.5
             activerecord: 52
+          # EOL 2023
           - ruby: 2.7
             activerecord: 60
+          # rails 6.1 and 7.0 have different ruby versions
           - ruby: 2.7
             activerecord: 61
           - ruby: "3.0"
             activerecord: 61
-          - ruby: 3.1
-            activerecord: 70
-          - ruby: 3.2
-            activerecord: 70
     env:
       # for the pg cli (psql, pg_isready) and possibly rails
       PGHOST: 127.0.0.1 # container is mapping it locally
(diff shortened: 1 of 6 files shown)


kbrock:
@kshnurov could you please share the use case where this is running the same tests twice?

kbrock:
@kshnurov I am holding off merging this because you expressed dissatisfaction with this PR. But I don't know what your issue is.

These changes are useful in helping me diagnose problems with counter caching and depth caching. (an issue that you asked me to look into)

Please share what your issue is and I will do my best to resolve it.

kshnurov:
> I hadn't noticed that https://github.com/stefankroes/ancestry/pull/597 dropped all the materialized_path testing.

Because it didn't drop anything, it's still there. See [#622](https://github.com/stefankroes/ancestry/issues/622#issuecomment-1464869861)

kbrock:
Paraphrasing @kshnurov comments and reformatted a little from another thread. Best to have a conversation in one place:

>> I am adding back the backwards compatible tests that you dropped in https://github.com/stefankroes/ancestry/pull/597.

> [...it is still there...], I just renamed `STRATEGY` to `FORMAT` in https://github.com/stefankroes/ancestry/pull/597.
[It is testing](https://github.com/stefankroes/ancestry/pull/597/files#diff-c28d998df09587b54183f818b02267135199831921115fc3870790dd194b2a77R80-R81) both strategies, without `FORMAT` it'll be the default `materialized_path`.

```bash
bundle exec rake
FORMAT=materialized_path2 bundle exec rake
```

> You've added format to matrix in https://github.com/stefankroes/ancestry/pull/624, which is really good, but you didn't remove the first `bundle exec rake`, and now all tests [are run twice](https://github.com/stefankroes/ancestry/actions/runs/4390484813/jobs/7688961325). [... there are ...] two consecutive `bundle exec rake` [here](https://github.com/stefankroes/ancestry/pull/624/files#diff-c28d998df09587b54183f818b02267135199831921115fc3870790dd194b2a77R86-R87).

suggestion:

```diff
- bundle exec rake
  FORMAT=#{{ matrix.format }} bundle exec rake
```

---

Thank you for finding that.

kbrock:
update:
- removed duplicate rake
- fixed testing so all versions of ruby are tested
- testing postgres with both update strategies.

@kshnurov I was thinking about moving the format back into the single test run. This time into their own sections. It will be well documented like we do with databases.

Why:

The container setup overshadows the time it takes to run the tests. Things like obtaining a container adds additional hidden overhead. So while each test run is maybe 30 seconds faster,  there are double the number of runs - so the overhead is much more pronounced.

@kshnurov It sounded like you liked the matrix addition, so I wanted to ask if you had a strong opinion if I moved those back into the individual runs?

kbrock:
update:
- fixed heading for mysql test

ok. This seems to run very quickly right now.
If it poses a problem in the future, will possibly merge them together in the future

PR #626, opened 2023-03-11
Deprecate adding ancestry to the middle of an STI tree

kbrock:
This PR is also being merged into 4.3.x branch via #625

Unsure if this will be dropped in 5.0 or 6.0 but setting the wheels in motion.
But either way, we need to fix the test so that it actually tests what it set out to test.

### DIFF
--- a/.github/workflows/run_test_suite.yml
+++ b/.github/workflows/run_test_suite.yml
@@ -1,9 +1,9 @@
 name: run-test-suite
 on:
   push:
-    branches: [ master ]
+    branches: [ master, 4-3-stable ]
   pull_request:
-    branches: [ master ]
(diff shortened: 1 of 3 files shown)


kbrock:
as hashed out in #620 we will keep this column for now. We can revisit at a later time to see if having all the `base_ancesty_class` references buys us something over the standard rails `base_class`

kshnurov:
What's the purpose of this deprecation? Makes zero sense.

kbrock:
Agreed, I am putting together a PR that removes this deprecation and adds more tests on this case. Am waiting for 600 / 620 to go in.

I still wonder about the need for `base_ancestry_class`, but can address that at a later time.

I also need to remove the deprecation from 4-3-stable.

PR #627, opened 2023-03-11
Root node has no siblings.

### LINKED ISSUE #590: Multi-tenant with multiple roots issue calling siblings
We have a scenario where we have a multi-tenanted application with each tenant having a set of roots with children (i.e. no singular root node).

Is it possible to scope all of the ancestry queries to a single-tenant as we're currently seeing calls to `root.siblings` crossing the tenant boundary?

I'd rather not introduce an unnecessary super-root type from which our client 'roots' descend, since we'd have to special case a lot of our tree logic. I'd also rather not introduce a default_scope to the model. Is there another approach?

Cheers!

kbrock:
This is a breaking change.

Each tree is a unique tree, so the root node of a tree has no siblings.

Fixes #590

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ The yellow nodes are those returned by the method.
 
 When using `STI` all classes are returned from the scopes unless you specify otherwise using `where(:type => "ChildClass")`.
 
-<sup id="fn1">1. [other root records are considered siblings]<a href="#ref1" title="Jump back to footnote 1.">↩</a></sup>
+<sup id="fn1">1. [other root records are not considered siblings]<a href="#ref1" title="Jump back to footnote 1.">↩</a></sup>
 
 # has_ancestry options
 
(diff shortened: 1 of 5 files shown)


kbrock:
@kshnurov do you feel that a root node should have all other root nodes as siblings?

Would you prefer that this was behind a feature switch?
Seems to me that the current behavior is a bug.

kshnurov:
> @kshnurov do you feel that a root node should have all other root nodes as siblings?

Yes, it should, I might want to get other root nodes without having to do `where(ancestry: self.class.root_ancestry)`, where `root_ancestry` is an internal gem method not even described in README.

#590 can just do `root? ? nil : siblings` without having all users of this gem to rewrite their code because of the breaking change no one really asked for.

kbrock:
To find roots you just use `Model.roots`. No internal methods required.

Looks like `closure_trees` has `root.siblings` point to all roots as well.
I will tell #590 how to solve his problem.

closing

PR #629, opened 2023-03-13
Ensure has_ancestry in subclasses do not leak

kbrock:
- BaseClass
  - ChildClass1 (defined `has_ancestry`)
  - ChildClass2 (also defines `has_ancestry`)

Ensure that `ChildClass1` ancestry options do not leak over to `ChildClass2`
If there are multiple STI nodes that have ancestry defined.

TODO: after merging #628 rewrite this PR to ensure `ancestry_options` of the two trees are different (instead of the deprecated counter_cache_column).

TODO: Do we want to un-deprecate `has_ancestry` in the middle of the tree?


kshnurov:
That's one more thing a #620 would break.

kbrock:
> That's one more thing a #620 would break.

Yup. I tried to think of any issues we could have in STI trees. This test was suspect and it got me thinking of other places bugs may lurk. `@@variables` are notorious for introducing bugs within hierarchies, let alone STI trees.

PR #630, opened 2023-03-13
More sti tests 43

kbrock:
backport sti tests to 4-3-stable
un-deprecate has_ancestry not from the root of an STI tree


kbrock:
This should have targeted 43, but targeted master. see #631

PR #633, opened 2023-03-14
update Ancestry writers

kbrock:
extracted from #600 

These are now non-editable class level variables:
- `ancestry_base_class`
- `ancestry_column`
- `ancestry_delimiter` (recently introduced #597 )

If this breaks in developer code try the class variable e.g.: `self.class.ancestry_column`


kbrock:
update:
- found one more spot accessing instance variable `ancestry_base_class`

PR #637, opened 2023-03-16
[4-3] add back ancestor_ids_in_database

### LINKED ISSUE #636: ancestor_ids_in_database removed?
ancestry (4.2.0) has `ancestor_ids_in_database`, but ancestry (4.3.0) doesn't.

[This pull request](https://github.com/stefankroes/ancestry/pull/589) removes `ancestor_ids_in_database`.

Is this intentional?

If this is intentional, what is alternative?

kbrock:
this was removed by #589

adding fields back.
added tests to exercise

Also changed behavior of children, descendants, and indirects:
- was using ancestry_in_database, now using ancestry

This does not cause a change in our orphan_strategy changes, nor our hook tests, but theoretically this may cause differences when traversing the ancestry tree after a major change.

meaning: it should result in better behavior

Fixes #636 /cc @Znz

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 Doing our best at supporting [SemVer](http://semver.org/) with
 a nice looking [Changelog](http://keepachangelog.com).
 
+## Version [4.3.1] <sub><sup>2023-03-17</sub></sup>
+* Fix: added back fields that were removed in #589 [#637](https://github.com/stefankroes/ancestry/pull/637) (thx @znz)
+  - ancestor_ids_in_database
+  - ancestor_id_before_last_save
+  - parent_id_in_database
+  - parent_id_before_last_save
+
 ## Version [4.3.0] <sub><sup>2023-03-09</sub></sup>
 
 * Fix: materialized_path2 strategy [#597](https://github.com/stefankroes/ancestry/pull/597) (thx @kshnurov)
--- a/lib/ancestry/materialized_path.rb
+++ b/lib/ancestry/materialized_path.rb
@@ -124,8 +124,16 @@ module Ancestry
         parse_ancestry_column(read_attribute(self.ancestry_base_class.ancestry_column))
       end
 
+      def ancestor_ids_in_database
+        parse_ancestry_column(attribute_in_database(self.class.ancestry_column))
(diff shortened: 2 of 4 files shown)


kbrock:
update:
- fixed ancestry_in_database references
update:
- updated changelog

kbrock:
update:
- ugh, backport rebase flub. fixing descendant_before_save_conditions

kbrock:
this pr got too big.
It depended upon a number of tests
I may need to backport my spec changes and then possibly introduce 2 changes.

znz:
I tried with `gem 'ancestry', git: 'https://github.com/kbrock/ancestry.git', branch: 'add_ancestor_ids_in_database_43'`, but some tests failed.

After setting `parent_id`, `descendants` becomes empty. I expect `descendants` does not change.

```ruby
# foo has some children, and bar is in another tree.
foo.descendants # => some children
child = foo.descendants.first
foo.parent_id = bar.id
foo.descendants # => empty (expect some children)
child.parent == name_2 # => true
```

kbrock:
@znz Are you able to add a test case that will fail?

- what rails, ruby, and database are you running?

Maybe you can see how I am testing in #638 and will have better luck than I do with this descendant issue.

If you are able to help me write a test, then I can properly track down the issue and changes in 4-2 and 4-3

Thanks

kbrock:
found the culprit to the problem.

- Instead of cherry-picking tests into this PR, I backported a number of test PRs.
- I added a separate set of tests that exercise the descendants to make sure that issue does not get introduced.
- I have another thread around the differences between ancestry_ids_before_last_save

I am targeting 4.3.1, but I do wonder if 5.0 (master) would works for you. (assuming ancestor_ids_before_save works)

kbrock:
you can find this on stefankroes/ancestry `4-3-stable` or `master`

```ruby
gem 'ancestry', git: 'https://github.com/stefankroes/ancestry.git', branch: '4-3-stable'
```

let me know if you have any other problems

znz:
Thanks.

`4-3-stable` branch works well.

kbrock:
@znz Hope 4.3.1 works for you.
Thank you for reporting the issue and helping test this

PR #649, opened 2023-03-25
Drop touch and ancestry_primary_key_format

kbrock:
These parameters are passed into `has_ancestry` and consumed by it.

This is part of a trend to drop many of the exposed/class level configuration parameters in ancestry.

# touch

### Before

we referenced touch to determine if we should touch fields

### After

We do not install the callbacks unless we want the touch behavior.
Variable is now no longer needed.

# ancestry_primary_key_format

### Before

This was recently introduced in 3.4 and is a mechanism for passing the key format into the format class (e.g.: `materialized_path`) to determine the regular expression used for validation

### After

It is being passed into the format and no longer referenced via the class level variable. It no longer needs to be an exposed.

### DIFF
--- a/lib/ancestry/has_ancestry.rb
+++ b/lib/ancestry/has_ancestry.rb
@@ -30,6 +30,7 @@ module Ancestry
       cattr_reader :ancestry_base_class, instance_reader: false
 
       # Touch ancestors after updating
+      # days are limited. need to handle touch in pg case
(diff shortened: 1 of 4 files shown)


kbrock:
updates:

Since we were modifying touch, I added the tests here
Looks like this properly highlights an issue here. Will pull that out so we can apply to 3-4 stable.

PR #650, opened 2023-03-25
[4-3] fix touch tests

kbrock:
some tests were not reloading the records to pickup changes
some tests were starting with a recent updated_at. so they always showed as changed

This helps expose that there is a problem with `update_strategy == :sql`

Updating hook and touch tests to fix this common problem.
touch with mysql on materialized_path also showed some strange behavior. Think it is isolated to tests as we tend to not change the update_at into the past out from underneath a record in production.


kbrock:
these tests are failing on mysql for materialized_path
other database working
other formats working

PR #656, opened 2023-03-28
Fixes `.sort_by_ancestry` with custom `ancestry_column`

mitsuru:
Fixes  `NoMethodError: undefined method `ancestry'` in `.sort_by_ancestry`

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -98,7 +98,7 @@ module Ancestry
 
       unless arranged
         presorted_nodes = nodes.sort do |a, b|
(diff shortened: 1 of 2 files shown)


kbrock:
2 parts are not well tested:
- using non `id` for the primary key. I think rails is pretty good at translating this for us (in `read_attribute` behind the scenes)
- using a column other than `ancestry`. These spot tests like you introduced may help a little, but we need to probably bite the bullet and create a suite that runs for a totally different ancestry column.

kbrock:
Thank you.
I'm going to merge this and then introduce tests that ensure we are testing alternative ancestry columns for the whole suite

PR #658, opened 2023-03-28
Extensible orphan_strategy

### LINKED ISSUE #142: Custom orphan_strategy would be desirable
Hello,

I'm trying to rearrange the children of a node in a before_destroy filter.

Suppose the following structure:

n1
|-n2
|-n3
|-n4

n1 has three children, all siblings.

After destroying n1, I would like n2 to became the parent of n3 and n4:

n2
|-n3
|-n4

Since any of the orphan strategies meets my requirements, I need a way to implement it myself. I tried to do that with a before_destroy filter, but it seems that it is not possible.

I'm trying to set n2 parent to nil, and n3 and n4 parents to n2, but it seems that the children collection is cached and the orphan strategy is run over the cached children collection.

Am I missing anything? Is there a way to implement a custom orphan strategy?

kbrock:
This introduces 2 ways for developers to extend the built in orphan strategies.

Fixes #142 

/cc @brendon Here is custom orphan strategies.

After implementing this, I'm thinking of keeping `:none` and dropping the custom string.

- `none` is straightforward and simple to document. The custom one is much harder (red flag)
- There is a gotcha requiring them to define it first. Requires a bunch of documentation (red flag)
- Defining your own `before_destroy` seems simple.
- Avoiding the build the callback gook would be nice. Maybe leaving this code as is but not documenting it may be a plan as well.
- I don't like people knowing about and overriding `apply_orphan_strategy`, but maybe it is not so bad.

As a developer, what is your take?

---

## Turning orphan strategy handling off

```ruby
class ModelBefore
  def apply_orphan_strategy # no longer necessary with :none
    # skip
  end
  has_ancestry orphan_strategy: :none
end

class ModelAfter
  has_ancestry orphan_strategy: :none
end
```

## Implementing a custom orphan strategy

```ruby
module CustomStrategy
  def apply_orphan_strategy_abc
    # custom goodness
  end
end

class ModelAfter1
  includes CustomStrategy
  has_ancestry orphan_strategy: :abc
end

class ModelAfter2
  includes CustomStrategy
  before_destroy :apply_orphan_strategy_abc
  has_ancestry orphan_strategy: :none
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,33 +164,34 @@ The `has_ancestry` method supports the following options:
     :ancestry_column       Column name to store ancestry
                            'ancestry' (default)
     :ancestry_format       Format for ancestry column (see Ancestry Formats section):
-                           :materialized_path (default)     1/2/3, root nodes ancestry=nil
-                           :materialized_path2 (preferred) /1/2/3/, root nodes ancestry=/
-    :orphan_strategy       Instruct Ancestry what to do with children of a node that is destroyed:
+                           :materialized_path   1/2/3, root nodes ancestry=nil (default)
+                           :materialized_path2  /1/2/3/, root nodes ancestry=/ (preferred)
+    :orphan_strategy       How to handle children of a destroyed node:
                            :destroy   All children are destroyed as well (default)
                            :rootify   The children of the destroyed node become root nodes
                            :restrict  An AncestryException is raised if any children exist
                            :adopt     The orphan subtree is added to the parent of the deleted node
                                       If the deleted node is Root, then rootify the orphan subtree
-    :cache_depth           Cache the depth of each node (See Depth Cache section)
+    :cache_depth           Cache the depth of each node: (See Depth Cache section)
                            false   Do not cache depth (default)
                            true    Cache depth in 'ancestry_depth'
-                           String  Cache depth in column referenced
-    :primary_key_format    regular expression that matches the format of the primary key
-                           '[0-9]+'           (default) integer ids
-                           '[-A-Fa-f0-9]{36}'           UUIDs
-    :touch                 Instruct Ancestry to touch the ancestors of a node when it changes
-                           false (default) don't invalide nested key-based caches
-    :counter_cache         Whether to create counter cache column accessor.
-                           false (default) don't store a counter cache
-                           true            store counter cache in `children_count`.
-                           String          name of column to store counter cache.
-    :update_strategy       Choose the strategy to update descendants nodes
-                           :ruby (default) All descendants are updated using the ruby algorithm.
-                                           This triggers update callbacks for each descendant node
-                           :sql            All descendants are updated using a single SQL statement.
-                                           This strategy does not trigger update callbacks for the descendants.
-                                           This strategy is available only for PostgreSql implementations
+                           String  Cache depth in the column referenced
+    :primary_key_format    Regular expression that matches the format of the primary key:
+                           '[0-9]+'            integer ids (default)
+                           '[-A-Fa-f0-9]{36}'  UUIDs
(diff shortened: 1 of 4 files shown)


lib/ancestry/has_ancestry.rb
```diff
@@ -60,19 +60,13 @@ def has_ancestry options = {}
...
-      when :adopt
-        alias_method :apply_orphan_strategy, :apply_orphan_strategy_adopt
-      when :restrict
-        alias_method :apply_orphan_strategy, :apply_orphan_strategy_restrict
-      else
+      orphan_strategy_helper = "apply_orphan_strategy_#{orphan_strategy}"
+      if method_defined?(orphan_strategy_helper)
+        alias_method :apply_orphan_strategy, orphan_strategy_helper
+        before_destroy :apply_orphan_strategy
+      elsif orphan_strategy.to_s != "none"
```
brendon, line 67: Could you just compare with the symbol to avoid calling `to_s`?

README.md
```diff
@@ -162,33 +162,36 @@ The `has_ancestry` method supports the following options:
...
+                           :materialized_path   1/2/3, root nodes ancestry=nil (default)
+                           :materialized_path2  /1/2/3/, root nodes ancestry=/ (preferred)
     :orphan_strategy       Instruct Ancestry what to do with children of a node that is destroyed:
                            :destroy   All children are destroyed as well (default)
                            :rootify   The children of the destroyed node become root nodes
                            :restrict  An AncestryException is raised if any children exist
                            :adopt     The orphan subtree is added to the parent of the deleted node
                                       If the deleted node is Root, then rootify the orphan subtree
+                           :none      skip this logic. (add your own `before_destroy`)
+                           String     Custom. `before_destroy :apply_orphan_strategy_#{String}` to handle orphans.
```
brendon, line 174: This might need a little expansion so that people know they don't need to add their own `before_destroy` block here. Perhaps:

Your custom destroy strategy method named with the convention `apply_orphan_strategy_#{String}` will be executed as an `before_destroy` callback.

brendon (changes requested):
Looks pretty good. Just a couple of suggestions there :D

README.md
kbrock: Yea, After trying to document this had me wanting to drop the feature `apply_orphan_strategy_#{string}` and just have the person write a `before_destroy`

brendon: Yea that makes sense :)

kbrock:
Ok, so this stalled.

I think that introducing `none` makes sense while introducing a naming strategy is too much.
The issues with making sure the method is defined first and the difficulty in documentation makes me feel that this is not ready yet.

I will keep this PR intact, and create another PR to introduce none

kbrock:
update:
- handle alternate ancestry column names in custom ancestry test

brendon:
Looks good :) When I'm upgrading next, I'll give it a go. It might be that the custom orphan strategy would need to be defined immediately after `has_ancestry` to ensure callbacks are still in the correct order in some apps :)

kbrock:
Actually, the custom method needs to be defined before the `has_ancestry`.
So just adding an `include` would work.

But in looking through github, the overrides for this method were mostly to disable it.
There was one that moved this to a background task. (was thinking of changing these methods to pass an object, but the reliance on changed makes it tricky)

brendon:
I think in my case it was the order of deletion but I think that was also fixed recently so I might not need the custom method anyway. I've not started moving to Rails 7 yet but when I do, this gem will get upgraded :D

kbrock:
@brendon Thanks. That info helps me.

Someone asked about when this will ship. I'm probably going to backport this, as I am making (slow) progress towards converting many functions to sql only.

PR #659, opened 2023-03-29
Alt ancestry column suite

kbrock:
This is followup from #656

Running test suite with an alternate ancestry column to ensure all tests work with an alt ancestry column.

@mitsuru These tests should be more complete for you.


mitsuru:
@kbrock 
For me this gem is one of the indispensable ones and I am always grateful for it.

PR #661, opened 2023-03-29
[WIP] Add option on how to handle orphans when arranging nodes

### LINKED ISSUE #539: arrange that discards orphans
Hi @kbrock,

I was wondering if there was appetite for altering the `arrange` method to allow for an option to drop orphans instead of promoting them to root nodes?

The use case is for when we select a part of a tree, and then also have conditions that remove some intermediate nodes but not their descendants.

kbrock:
Fixes #539

@brendon is this what you had in mind?
How do you usually get into this method?

read: I'm wondering where we want to pass `add_orphans: false`


lib/ancestry/class_methods.rb
```diff
@@ -40,14 +40,14 @@ def arrange options = {}
     # @param nodes [Array[Node]] nodes to be arranged
     # @returns Hash{Node => {Node => {}, Node => {}}}
     # If a node's parent is not included, the node will be included as if it is a top level node
-    def arrange_nodes(nodes)
+    def arrange_nodes(nodes, add_orphans: true)
```
brendon, line 43: It might make more sense to call is `rootify_orphans` or perhaps in the reverse `discard_orphans`?

brendon (changes requested):
Looks good. Would definitely need a test or two to ensure it's working as it's such a mind-job every time reading that code :D

kbrock:
WIP: functionality complete. need tests

kbrock:
rebased
more tests

PR #665, opened 2023-05-29
fix: `rebuild_counter_cache!` primary key cast problem

RongRongTeng:
Fix 
- https://github.com/stefankroes/ancestry/pull/663

Currently, when the primary key consists of more than one character and casts to `CHAR` without a specific length, it can lead to problems. This occurs because the default length for CHAR is 1, which may not be sufficient to accommodate the values of the primary key.

To resolve this issue, this PR changes casting the primary key column to `VARCHAR`.

### DIFF
--- a/lib/ancestry/materialized_path.rb
+++ b/lib/ancestry/materialized_path.rb
@@ -94,8 +94,8 @@ module Ancestry
 
     def child_ancestry_sql
       %{
-        CASE WHEN #{table_name}.#{ancestry_column} IS NULL THEN CAST(#{table_name}.#{primary_key} AS CHAR)
-        ELSE      #{concat("#{table_name}.#{ancestry_column}", "'#{ancestry_delimiter}'", "CAST(#{table_name}.#{primary_key} AS CHAR)")}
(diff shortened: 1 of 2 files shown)


kbrock:
this looks great. wonder why it did not run tests

kbrock:
kicking. trying to get the build to run

kbrock:
I have a few ideas to improve these tests.

looks like VARCHAR will not work with mysql, but I think I can get the casts using `concat()` or something like that.

I'll open a PR based upon your commits and ping you shortly

kbrock:
Please let me know if #668 meets your needs.

I gave you attribution for one of the commits but had to change it a bit. Please tell me if you are not ok with the changes I made to your code.

RongRongTeng:
> Please let me know if #668 meets your needs.
> 
> I gave you attribution for one of the commits but had to change it a bit. Please tell me if you are not ok with the changes I made to your code.

Thanks! It meets my needs  🙏🏻 This bug occurs when using PostgreSQL.
I will close this PR.

PR #667, opened 2023-07-12
README update: Change depth_cache: true to cache_depth: true

onerinas:
Using `depth_cache: true` given an error `Unknown option for has_ancestry: :depth_cache => true`

Correct option is `cache_depth: true`


kbrock:
thank you

PR #689, opened 2024-10-23
Drop rails <6.0

kbrock:
Not sure if we want to drop older rubies

rails 6.0 doesn't support ruby 3.0, so dropping 6.0 simplifies the matrix. It was EOL 2023
rails 6.1 was EOL 2024. I'm open to adding this back if someone needs, but the 7.0 code base changes a lot in the internals that we may want to follow.


kbrock:
@SleeplessByte Thanks. I knew there was a reason I hadn't merged this yet.

You proposing keeping 6.1. Opinion on 6.0 regards to Debian?
My team is having trouble upgrading 7.1 to 7.2. It is a definite balancing act between supporting viable versions and throwing out old cruft.

SleeplessByte:
Personally I keep my gems working on stable Debian but not maintenance Debian. So at this point I have dropped 6.0 even though on Debian it still gets security fixes.

I'm just one person 😁

SleeplessByte:
That would be fine with me tbh. Goed beyond what many gem maintainers do.

PR #693, opened 2025-02-08
Optimize has_siblings? to use exists?

a5-stable:
Greetings.
This PR optimizes the has_siblings? method to improve performance by using `exists?`
This aims to reduce unnecessary scan for sibling existence and improve the performance of sibling checks, especially when there are many records.

### Changes:
```SQL
# Before
SELECT COUNT(*) FROM "test_nodes" WHERE "test_nodes"."ancestry" IS X

# After
SELECT 1 AS one FROM "test_nodes" WHERE "test_nodes"."ancestry" IS X AND nodes"."id" != X
```

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -255,7 +255,7 @@ module Ancestry
     end
 
     def has_siblings?
-      siblings.count > 1
+      siblings.where.not(self.class.primary_key => id).exists?
     end
     alias_method :siblings?, :has_siblings?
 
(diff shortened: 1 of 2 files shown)


kbrock:
@a5-stable I merged #694 - could you rebase this onto the new `master`?

lib/ancestry/instance_methods.rb
```diff
@@ -255,7 +255,7 @@ def sibling_ids
     end
 
     def has_siblings?
-      siblings.count > 1
+      siblings.where.not(self.class.primary_key => id).exists?
```
kbrock, line 258: future request: wish there were a way to cache this somehow so we could fetch this answer in a single query. Along with has_children?

PR #697, opened 2025-04-15
feat: eager loading

mhenrixon:
I am currently working on improving the performance of the generated queries, I figured I can let you tag along and offer feedback along the way.

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -3,9 +3,15 @@
 module Ancestry
   module ClassMethods
     # Fetch tree node if necessary
+    # Supports single objects or arrays for batch loading
     def to_node(object)
       if object.is_a?(ancestry_base_class)
         object
+      elsif object.is_a?(Array)
+        # Batch load multiple nodes at once to avoid N+1 queries
+        ids = object.map { |obj| obj.try(primary_key) || obj }
+        nodes = unscoped_where { |scope| scope.where(primary_key => ids) }
+        object.map { |obj| nodes.find { |n| n.id.to_s == (obj.try(primary_key) || obj).to_s } }
       else
         unscoped_where { |scope| scope.find(object.try(primary_key) || object) }
       end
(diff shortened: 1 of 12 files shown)


kbrock:
In my mind, the reason why this doesn't work is because we are using custom methods/scopes instead of associations.

One area probably the most in the way is the `to_node` aspect. When a scope is passed in.

I've looked at some other mechanisms that would us to use standard rails associations, so `preload` and `requires` would work.

lib/ancestry/class_methods.rb
```diff
@@ -29,11 +35,26 @@ def scope_depth(depth_options, depth)
...
+      
+      if (includes_param = options.delete(:includes))
+        scope = scope.includes(includes_param)
+      end
+      
+      if (preload_param = options.delete(:preload))
+        scope = scope.preload(preload_param)
+      end
+      
+      if (eager_load_param = options.delete(:eager_load))
```
kbrock, line 50: not sure if we want to pull off of `arrange` or we want to pull off the scope coming in.

E.g., `targets.try(:includes_value)` (thought we'd probably use standard and not `try`)

mhenrixon:
> I've looked at some other mechanisms that would us to use standard rails associations, so preload and requires would work.

I'd be down for some pair programming on this if you ever have the time and interest. 

The way this PR stands didn't solve it when running live but I'll keep stabbing at it. 

I have a long term goal of solving this and decided to take a first stab at it. 

Initial attempt didn't yield much of a difference. 

No stress on this

kbrock:
Hey @mhenrixon 
I have figured out how to have children, parent, root as associations.
What specific use case were you solving.

I have a prototype of descendants, but it is a little monkeypatchy - basically modify rails so I can use a `LIKE` for the has_many association.

If you're still using this project, I'd love to hear about your needs and I could share what I'm doing. See if it aligns.

--

Just reread your last comment - parent_id and root_id as actual columns (or sql virtual columns) is working with relationships. Just curious if that works for you or you need others.

kbrock:
have eager loading for parent, children, and root.

eager loading for descendents is a little ways off. I have it working but looking to make it have less monkey patching.

hope all is well.

Issue #695, opened 2025-02-16
Allow customization of column for ancestry relationships (using non-primary key)


a5-stable:
Greetings,

The ancestry gem currently assumes that the parent-child relationship is managed using the primary key.
However, there are certain cases where a different unique column needs to be used to manage the relationship when we use other columns as a identifier.

This is how it would behave;
```ruby
class Category
   has_ancestry ancestry_identifier_column: :another_identifier_column
end

root = Category.create!(another_identifier_column: 100)
node = Category.new(another_identifier_column: 200)
node.ancestor_ids = [root.another_identifier_column]
node.save!

node.ancestry
#=> 100
```

Some gems allow customizing the column used for managing parent-child relationships. For instance, the [awesome_nested_set](https://github.com/collectiveidea/awesome_nested_set) gem provides a `primary_column` option that enables users to specify which column should be used as the parent reference (parent_id). 

I am currently working on a PR to add this feature to ancestry, as I need it for my product.  I’d be happy to submit it if you find this approach acceptable.

a5-stable:
@kbrock 
Thank you for your feedback!
I understand that this may feel like a big change, but the change itself can actually be straightforward.
I have drafted a sample PR that aims to abstract the id or primary_key specification as a starting point.
I would appreciate it if you could take a look when you have a moment.
https://github.com/stefankroes/ancestry/pull/698

After this, the next steps for this feature would be:
- Introduce an option to specify a custom `ancestry_identifier_column`
- Add `ancestry_identifier_column_key_format`

kbrock:
this looks like it will add a lot of `send`/`read_attribute` changes. and making the code less readable. It is possible that it will find some edge cases that don't work for custom ids.

I do wonder if the `id` field breaks for a different `primary_key` - but we run the whole test suite with that use case and it passes. (not saying it won't break, just saying we do make an honest try)

For some reason, those two initiatives feel like they overlap. Feels like enabling a different `primary_key` would be similar to enabling a different key all together.

Again, if you want to get this feature in, we need to find a way that this is not a whole bunch of extra with added `send` and `read_attribute` calls, and simple variable names that are now more complex.

PR #698, opened 2025-05-11
Abstract all primary key references

a5-stable:
Sample PR for #695 

（This pull request introduces a significant refactor to the `ancestry` library by replacing direct references to `primary_key` and `id` with a new method, `ancestry_identifier_column`）

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -7,7 +7,7 @@ module Ancestry
       if object.is_a?(ancestry_base_class)
         object
       else
-        unscoped_where { |scope| scope.find(object.try(primary_key) || object) }
+        unscoped_where { |scope| scope.find(object.try(ancestry_identifier_column) || object) }
       end
     end
 
@@ -43,11 +43,11 @@ module Ancestry
     # @returns Hash{Node => {Node => {}, Node => {}}}
     # If a node's parent is not included, the node will be included as if it is a top level node
     def arrange_nodes(nodes)
-      node_ids = Set.new(nodes.map(&:id))
+      node_ids = Set.new(nodes.map(&:ancestry_identifier_column))
       index = Hash.new { |h, k| h[k] = {} }
 
       nodes.each_with_object({}) do |node, arranged|
-        children = index[node.id]
+        children = index[node.ancestry_identifier_column]
         index[node.parent_id][node] = children
         arranged[node] = children unless node_ids.include?(node.parent_id)
       end
@@ -125,25 +125,25 @@ module Ancestry
           # ... check validity of ancestry column
           if !node.sane_ancestor_ids?
             raise Ancestry::AncestryIntegrityException, I18n.t("ancestry.invalid_ancestry_column",
-                                                               :node_id => node.id,
+                                                               :node_id => node.ancestry_identifier_column,
                                                                :ancestry_column => node.read_attribute(node.class.ancestry_column))
           end
           # ... check that all ancestors exist
           node.ancestor_ids.each do |ancestor_id|
             unless exists?(ancestor_id)
               raise Ancestry::AncestryIntegrityException, I18n.t("ancestry.reference_nonexistent_node",
-                                                                 :node_id => node.id,
+                                                                 :node_id => node.ancestry_identifier_column,
                                                                  :ancestor_id => ancestor_id)
             end
           end
           # ... check that all node parents are consistent with values observed earlier
-          node.path_ids.zip([nil] + node.path_ids).each do |node_id, parent_id|
-            parents[node_id] = parent_id unless parents.key?(node_id)
-            unless parents[node_id] == parent_id
+          node.path_ids.zip([nil] + node.path_ids).each do |node_identifier, parent_identifier|
+            parents[node_identifier] = parent_identifier unless parents.key?(node_identifier)
+            unless parents[node_identifier] == parent_identifier
               raise Ancestry::AncestryIntegrityException, I18n.t("ancestry.conflicting_parent_id",
-                                                                 :node_id => node_id,
-                                                                 :parent_id => parent_id || 'nil',
-                                                                 :expected => parents[node_id] || 'nil')
+                                                                 :node_id => node_identifier,
+                                                                 :parent_id => parent_identifier || 'nil',
+                                                                 :expected => parents[node_identifier] || 'nil')
             end
           end
         rescue Ancestry::AncestryIntegrityException => e
@@ -172,25 +172,25 @@ module Ancestry
               end
             end
             # ... save parent id of this node in parent_ids array if it exists
-            parent_ids[node.id] = node.parent_id if exists? node.parent_id
+            parent_ids[node.ancestry_identifier_column] = node.parent_id if exists? node.parent_id
 
             # Reset parent id in array to nil if it introduces a cycle
-            parent_id = parent_ids[node.id]
-            until parent_id.nil? || parent_id == node.id
-              parent_id = parent_ids[parent_id]
+            parent_identifier = parent_ids[node.ancestry_identifier_column]
+            until parent_identifier.nil? || parent_identifier == node.ancestry_identifier_column
+              parent_identifier = parent_ids[parent_identifier]
             end
-            parent_ids[node.id] = nil if parent_id == node.id
+            parent_ids[node.ancestry_identifier_column] = nil if parent_identifier == node.ancestry_identifier_column
           end
 
           # For each node ...
           scope.find_each do |node|
             # ... rebuild ancestry from parent_ids array
-            ancestor_ids, parent_id = [], parent_ids[node.id]
-            until parent_id.nil?
-              ancestor_ids, parent_id = [parent_id] + ancestor_ids, parent_ids[parent_id]
+            ancestor_identifiers, parent_identifier = [], parent_ids[node.ancestry_identifier_column]
+            until parent_identifier.nil?
+              ancestor_identifiers, parent_identifier = [parent_identifier] + ancestor_identifiers, parent_ids[parent_identifier]
             end
             node.without_ancestry_callbacks do
-              node.update_attribute :ancestor_ids, ancestor_ids
+              node.update_attribute :ancestor_ids, ancestor_identifiers
             end
           end
         end
@@ -204,7 +204,7 @@ module Ancestry
           node.without_ancestry_callbacks do
             node.update_attribute :ancestor_ids, ancestor_ids
           end
-          build_ancestry_from_parent_ids! column, node.id, ancestor_ids + [node.id]
+          build_ancestry_from_parent_ids! column, node.ancestry_identifier_column, ancestor_ids + [node.ancestry_identifier_column]
         end
       end
     end
@@ -233,11 +233,11 @@ module Ancestry
         connection.execute %{
           UPDATE #{table_name} AS dest
           LEFT JOIN (
-            SELECT #{table_name}.#{primary_key}, COUNT(*) AS child_count
+            SELECT #{table_name}.#{ancestry_identifier_column}, COUNT(*) AS child_count
             FROM #{table_name}
             JOIN #{table_name} children ON children.#{ancestry_column} = (#{child_ancestry_sql})
-            GROUP BY #{table_name}.#{primary_key}
-          ) src USING(#{primary_key})
+            GROUP BY #{table_name}.#{ancestry_identifier_column}
+          ) src USING(#{ancestry_identifier_column})
           SET dest.#{counter_cache_column} = COALESCE(src.child_count, 0)
         }
       else
@@ -262,6 +262,11 @@ module Ancestry
       else
         @primary_key_is_an_integer = !ANCESTRY_UNCAST_TYPES.include?(type_for_attribute(primary_key).type)
       end
+    end
+
+    # alias for primary_key
+    def ancestry_identifier_column
+      primary_key
     end
   end
 end
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -4,7 +4,7 @@ module Ancestry
   module InstanceMethods
     # Validate that the ancestors don't include itself
     def ancestry_exclude_self
-      errors.add(:base, I18n.t("ancestry.exclude_self", class_name: self.class.name.humanize)) if ancestor_ids.include?(id)
+      errors.add(:base, I18n.t("ancestry.exclude_self", class_name: self.class.name.humanize)) if ancestor_ids.include?(ancestry_identifier_column)
     end
 
     # Update descendants with new ancestry (after update)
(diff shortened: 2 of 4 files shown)


lib/ancestry/class_methods.rb
```diff
@@ -7,7 +7,7 @@ def to_node(object)
       if object.is_a?(ancestry_base_class)
         object
       else
-        unscoped_where { |scope| scope.find(object.try(primary_key) || object) }
+        unscoped_where { |scope| scope.find(object.try(ancestry_identifier_column) || object) }
```
kbrock, line 10: here, primary_key (`:id`), vs ancestry_identifier_column (`:id`) -- I'll keep them the same for now, but want to set this to find other issues

lib/ancestry/class_methods.rb
```diff
@@ -43,11 +43,11 @@ def arrange(options = {})
     # @returns Hash{Node => {Node => {}, Node => {}}}
     # If a node's parent is not included, the node will be included as if it is a top level node
     def arrange_nodes(nodes)
-      node_ids = Set.new(nodes.map(&:id))
+      node_ids = Set.new(nodes.map(&:ancestry_identifier_column))
       index = Hash.new { |h, k| h[k] = {} }
 
       nodes.each_with_object({}) do |node, arranged|
-        children = index[node.id]
+        children = index[node.ancestry_identifier_column]
```
kbrock, line 50:
```suggestion
        children = index[node.send(ancestry_identifier_column)]
```

lib/ancestry/class_methods.rb
```diff
@@ -125,25 +125,25 @@ def check_ancestry_integrity!(options = {})
           # ... check validity of ancestry column
           if !node.sane_ancestor_ids?
             raise Ancestry::AncestryIntegrityException, I18n.t("ancestry.invalid_ancestry_column",
-                                                               :node_id => node.id,
+                                                               :node_id => node.ancestry_identifier_column,
```
kbrock, line 128:
```suggestion
                                                               :node_id => node.read_attribute(ancestry_identifier_column),
```

lib/ancestry/class_methods.rb
```diff
@@ -125,25 +125,25 @@ def check_ancestry_integrity!(options = {})
...
                                                                  :ancestor_id => ancestor_id)
             end
           end
           # ... check that all node parents are consistent with values observed earlier
-          node.path_ids.zip([nil] + node.path_ids).each do |node_id, parent_id|
-            parents[node_id] = parent_id unless parents.key?(node_id)
-            unless parents[node_id] == parent_id
+          node.path_ids.zip([nil] + node.path_ids).each do |node_identifier, parent_identifier|
+            parents[node_identifier] = parent_identifier unless parents.key?(node_identifier)
+            unless parents[node_identifier] == parent_identifier
```
kbrock, line 142: maybe we should leave as `node_id, parent_id` -- otherwise this is pretty verbose. What do you think? (I may change my mind later. honestly asking your opinion)

lib/ancestry/class_methods.rb
kbrock: `id` is nice. behind the scenes it does smart stuff. wonder if something similar would work here.

kbrock:
Yea, this is getting confusing for me.

So parent_id is the id from the parent node, but that is technically the parent's ancestry_identifier_column. So when we have root_id and parent_id, then it is the root_identity field and the parent_identity field.

So all lookups of any nodes in the tree need to change.

I'm still thinking breadcrumbing may be the way you want.

Do you have an example of the code that won't work with the current model?

lib/ancestry/class_methods.rb
kbrock: and find inherently is a `find_by(:id => object.id)`
So this should read `find_by(ancestry_identifier_column => object.read_attribute(ancestry_identifier_column))`

As I'm thinking more about this, this getting unwieldy very quickly.
ods are pretty low.

By the way, this method could be defined as `def to_node(id_or_object)`. The `try` is saying "if this is an `Object`, then lets fetch the `id`, otherwise lets assume it is an `Integer` `id`", which works because ruby dropped the `Object#id` maybe 15 years ago.'

lib/ancestry/class_methods.rb
a5-stable: Based on your [suggestion](https://github.com/stefankroes/ancestry/pull/698#discussion_r2098023509), I introduced the method `ancestry_identifier_value`, which returns `read_attribute(ancestry_identifier_column)`.

And `id` in Rails method implicitly returns the primary key value, so this `ancestry_identifier_value` can serve a similar purpose if I'm correct.

By using this method, I can avoid repeatedly writing `read_attribute` or `send` whenever I need to fetch the value, which helps keep the code more readable.

lib/ancestry/class_methods.rb
a5-stable: Thank you for your comment. Here's how I'm currently thinking about this:

- I don't plan to change the naming of interface-level fields like parent_id or node_id.

- These names are compatible with existing behavior, and semantically they make sense as the identifiers used within the context of ancestry gem. 

For this reason, I’ve reverted instances of parent_identifier back to parent_id. I'd appreciate your thoughts on this direction.

kbrock:
@a5-stable would your need be met if we had `ancestry` with was `id/id`, but also had `identifier_ancestry` which had `identifier/identifier`?

a5-stable:
I sincerely appreciate your careful and patient review....!
Let me explain more about the background, API design, and related details.

### Usecase
One Example of the use case I’m considering involves a model that uses [Temporal Modeling](https://en.wikipedia.org/wiki/Temporal_database).

Temporal models store historical data, so multiple records can represent the lifecycle of a single "resource." For example, using a [bitemporal model](https://en.wikipedia.org/wiki/Bitemporal_modeling) might look like this:

```sql
CREATE TABLE employee_department_bitemporal (
    id               SERIAL PRIMARY KEY,       -- Surrogate Key
    bitemporal_id    INT NOT NULL,             -- Identifier for the logical employee
    valid_from       DATE NOT NULL,
    valid_to         DATE NOT NULL,
    transaction_from TIMESTAMP NOT NULL,
    transaction_to   TIMESTAMP NOT NULL,
    UNIQUE (bitemporal_id, valid_from, transaction_from)  
);
```
In this case, id is unique across each version of the record, while bitemporal_id uniquely identifies the logical employee.

Now, let’s say that we want to add a hierarchical association to this model. In some cases, it makes more sense for the ancestry column to store values like `bitemporal_id/bitemporal_id/...` instead of `id/id/...`, so that the hierarchy is consistent across the logical employees, not the history records.

> would your need be met if we had ancestry with was id/id, but also had identifier_ancestry which had identifier/identifier?

In the use case above, the coexistence strategy (having both id/id and identifier/identifier) might work, but I wonder if storing the unused id/id hierarchy leads to unnecessary code and data overhead.
Perhaps I am not fully appreciating the intention behind this strategy...

### public API such as `parent_id`
I understand that technically `parent_id` refers to the value in the parent's `ancestry_identifier_column`, which may differ from a simple primary key. In the context of the ancestry column, I think this is acceptable as an identifier, so it’s fine to keep familiar terms like `parent_id` and `root_id `in the public API.


### readability
Based on your https://github.com/stefankroes/ancestry/pull/698#discussion_r2098023509, I introduced the method ancestry_identifier_value, which returns read_attribute(ancestry_identifier_column).

By using this method, I think we can avoid repeatedly writing `read_attribute` or `send` whenever I need to fetch the value.

a5-stable:
@kbrock 
I sincerely appreciate your careful and patient review....!
Let me explain more about the background, API design, and related details.

### Usecase
One Example of the use case I’m considering involves a model that uses [Temporal Modeling](https://en.wikipedia.org/wiki/Temporal_database).

Temporal models store historical data, so multiple records can represent the lifecycle of a single "resource." For example, using a [bitemporal model](https://en.wikipedia.org/wiki/Bitemporal_modeling) might look like this:

```sql
CREATE TABLE employee_department_bitemporal (
    id               SERIAL PRIMARY KEY,       -- Surrogate Key
    bitemporal_id    INT NOT NULL,             -- Identifier for the logical employee
    valid_from       DATE NOT NULL,
    valid_to         DATE NOT NULL,
    transaction_from TIMESTAMP NOT NULL,
    transaction_to   TIMESTAMP NOT NULL,
    UNIQUE (bitemporal_id, valid_from, transaction_from)  
);
```
In this case, id is unique across each version of the record, while bitemporal_id uniquely identifies the logical employee.

Now, let’s say that we want to add a hierarchical association to this model. In some cases, it makes more sense for the ancestry column to store values like `bitemporal_id/bitemporal_id/...` instead of `id/id/...`, so that the hierarchy is consistent across the logical employees, not the history records.

> would your need be met if we had ancestry with was id/id, but also had identifier_ancestry which had identifier/identifier?

In the use case above, the coexistence strategy (having both id/id and identifier/identifier) might work, but I wonder if storing the unused id/id hierarchy leads to unnecessary code and data overhead.
Perhaps I am not fully appreciating the intention behind this strategy...

### public API such as `parent_id`
I understand that technically `parent_id` refers to the value in the parent's `ancestry_identifier_column`, which may differ from a simple primary key. In the context of the ancestry column, I think this is acceptable as an identifier, so it’s fine to keep familiar terms like `parent_id` and `root_id `in the public API.


### readability
Based on your https://github.com/stefankroes/ancestry/pull/698#discussion_r2098023509, I introduced the method ancestry_identifier_value, which returns read_attribute(ancestry_identifier_column).

By using this method, I think we can avoid repeatedly writing `read_attribute` or `send` whenever I need to fetch the value.

kbrock:
Ok, cool concept.

So basically the `id` field is not your primary key, but rather the `bitemporal_id` field.
And it is not unique, but rather only unique when giving a date in a 'valid_from..valid_to` range and probably assuming that `transaction_to` is `null`.

TANGENT: I'd check if creating your indexes with `WHERE transaction_to IS NULL` give you a good boost. (except for the case where you are trying to see what was returned in the past - which feels like an edge case)

So if you used `:id` as the ancestry key, then the tree would not be valid across a `valid` date boundary.

But also, just saying `parent_id` is not good enough, as you would need to specify the `valid` date (assuming that `transaction` date was `null`). But you would definitely want `parent_id` to be of the `bitemoral_id` type.

If you are recording a hierarchy in a point in time, then you want `ancestry` to be recorded at a point in time. Which seems to work, and would probably work better than closure trees. (not taking off the table, but seems to be easier to version using the ancestry path style)

When you move a node in a tree, you need to update a bunch of records. But in your case, you don't really want to modify any records (other than updating the transaction_to and possibly the valid_to dates). You'll need to ensure that updates in the callback do the necessary modification. Tricky edge case. Hopefully the update strategy system would work for you.

TANGENT: I'd like to moving towards sql transactions rather than updating the records in ruby and sending those across N+1 style. But time has been limited so that may never occur. But I think it may affect you.

TANGENT: Are you writing `UPDATE` triggers in the database to allow `UPDATE` statements but have those actually get converted to `INSERT` statements? May cause chaos for rails, or may make it more elegant.

TANGENT: Wondering if thread local variables [ref](https://www.rubytapas.com/2016/11/20/ruby-thread-local-variables/) for `transaction_time` (default to `null`) and a default `valid_time` would work. Then you could keep using the database as is with a default scope pulling the values out of the threaded variables. Not sure if you could go so far as to set the primary key to `bitemoral_id`, but maybe?

All the `find` operations need to change. so every scope and update in a callback would be wrong. This is a big change.

I agree that you don't want `id/id` because those are only valid at a single point in time, and  in the future, when you update the path, they may be invalid since the ids may span `transaction_time` scopes.

TANGENT: look into how the multi tenant solutions solve this scoping issue since the behavior may be similar to what you want. I notice that [multi-tenant gem](https://github.com/citusdata/activerecord-multi-tenant/blob/master/lib/activerecord-multi-tenant/multi_tenant.rb) uses the active support wrapper over thread local while [acts_as_tenant](https://github.com/ErwinM/acts_as_tenant/blob/master/lib/acts_as_tenant.rb) uses `Thread.current`.

Do keep in mind, in order to work with STI, the code does throw away default scopes, specifically when fetching a parent, but possibly in other cases, as well. So this may introduce issues depending upon how you implement your solution. A number of people over the years seem to be polar pro dropping scopes vs anti dropping scopes.

Ok, that was a little bit of a ramble.
This seems like it would require quite a bit of change. Need some time to process. Adding the `transaction` axis and the `valid` axis makes this quite interesting. For your domain do you need both, or would `transaction` date meet your needs? If the latter, it may be able to

kbrock:
ooh, I've used `deleted_at` for soft deletes and [paper_trail](https://rubygems.org/gems/paper_trail) for history. But this is a neat idea - though does look potentially intrusive

a5-stable:
@kbrock 
Thank you so much for your thoughtful and in-depth feedback. I really appreciate the time you took to consider this!
Your points are very helpful!

In my domain, we do require both the transaction and valid time axes—not just for logical deletion like discard, but to properly track the lifecycle of a resource as a historical record. (even in unitemporal use cases, we might want `bitemporal_id/bitemporal_id` as the ancestry key instead of id, to represent logical relationships rather than historical versions).

Also, I’m planning to use a multi-tenant-like approach, perhaps with something like the multi-tenant gem to manage valid and transaction times via thread-local variables. Your thoughts on default scopes and STI behavior were particularly helpful—I'll take extra care there.

Just to clarify, I shared this use case mainly to help illustrate the idea of supporting custom identifier columns for ancestry—not to require the entire bitemporal model to be supported out of the box.

I completely understand that this could lead to significant internal changes, and I’d be happy to help. If you have thoughts on where such a change should start or what kind of interface would make sense, I’d love to hear them and contribute.

Again, thank you for your time and insights.

kbrock:
> even in unitemporal use cases, we might want `bitemporal_id/bitemporal_id` as the ancestry key instead of `id`, to represent logical relationships rather than historical versions.

Yea, in your case, I feel `id/id` is not useful. Once you assign any `transaction_to` or valid_to`, then the `ancestry` value will be off.

a5-stable:
@kbrock 
Thank you very much for your thoughtful feedback and the time you’ve spent on this!
I fully understand that addressing this properly will require some time and effort, and I’m very willing to collaborate closely to move it forward.
If you have any concerns, suggestions, or ideas about where to begin or how to proceed, I would greatly appreciate your guidance.

kbrock:
@a5-stable I am not sure if this is related or totally tangental, and it is still an idea and not a know commodity, but...

```diff
- def parent_id
-   ancestry.split("/").last
- end
+ def parent_id
+   self.class.calculate_parent_id(self, ancestry_column)
+ end
+ def self.calculate_parent_id(rec, col_name)
+   rec.send(col_name).split("/").last
+ end
```

And further down the line, `parent_id` will not be part of a module, but dynamically generated in `has_ancestry`, much like how rails defines an attribute.

Does that change the way this PR is implemented?

kbrock:
And we have more.
The pattern I'm following is to define lightweight methods dynamically, building the methods in the model, and then having all the core functionality implemented in the implementation modules - since those are easier to debug, they have all the meat.

I've gotten all the primary key code in the generated module to remove assumptions around the primary key.

But the ones in the class/instance methods, those still need some fixing.
But we're getting there.

PR #700, opened 2025-06-25
Humanise model name for error message

### LINKED ISSUE #699: Class names are not properly humanized
The following PR introduced for model name: https://github.com/stefankroes/ancestry/pull/487
But the model name `ProductType` produces `Producttype`, not `Product type` as suggested in the PR.
EDIT: Not a bug. Fixed in new version.

labeebklatif:
Fixes #699 

The model name humanization was added in https://github.com/stefankroes/ancestry/pull/487 and the change was lost in internationalization attempt: https://github.com/stefankroes/ancestry/commit/f11cf9130a3b630e9d329fec2c33c8baf492c7d6. This PR adds it back.

> Imagine a model called ProductType. The old message was:
> `Producttype cannot be a descendant of itself.`
> The new message is:
> `Product type cannot be a descendant of itself.`

### DIFF
--- a/lib/ancestry/instance_methods.rb
+++ b/lib/ancestry/instance_methods.rb
@@ -4,7 +4,7 @@ module Ancestry
   module InstanceMethods
     # Validate that the ancestors don't include itself
     def ancestry_exclude_self
-      errors.add(:base, I18n.t("ancestry.exclude_self", class_name: self.class.name.humanize)) if ancestor_ids.include?(id)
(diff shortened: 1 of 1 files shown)


kbrock:
@labeebklatif Thank you.

Do you have ideas on adding a test so we don't introduce this again?

I'm not sure how to ensure that at test works on an English speaker's computer and fails on a computer with a different locale/language.

kbrock:
ok. Merging this as a fix. If you come up with some specs to prevent regressions, please put in another PR.

PR #729, opened 2026-03-13
Introduce MaterializedPath using ltree

kbrock:


sdhull:
Hi! 👋 thanks for your work on this—do you have a plan for when you'll cut a new release to include `ltree` support?

kbrock:
@sdhull so this introduced a bunch of new code. And it is complicated. Hoping within 2 weeks to clean up the builder and the loose end and release 6.0

6.0 will be a big release. Besides adding a bunch of serialization changes, the infrastructure has changed a bit - working towards allowing people to name these columns any way they want
7.0 will probably allow multiple ancestry columns per record (it is often requested but I feel it will not give people what they ask for)

The goal is to get this working like a standard rails association/has_many style with pk, fk, and association.

um. ltree is postgres only though. (are you running mysql?)

sdhull:
@kbrock we're running postgres! 
> Hoping within 2 weeks

Awesome! I'm pinning to latest commit on master branch for now 😬 😓 ... looking forward to updating gemfile with 6.0 😄 
> The goal is to get this working like a standard rails association/has_many style

that sounds appealing!

PR #747, opened 2026-03-28
Allow ancestry to define a primary ancestry key

### LINKED ISSUE #123: New release?
Any chance of a new release being made soon?  We'd quite like to use the new adopt orphan strategy.

Thanks!

kbrock:
## Description

Not sure if I want to commit, but putting this out there.

Many of the concepts are good. But not as readable as I'd like.
 A lot of change, so this will get stale quickly.

Fixes https://github.com/stefankroes/ancestry/pull/698

This allows ancestry to have `a/b/c/` but the primary keys be `1,2,3`

## Type of Change
- [ ] Feature

## Checklist
- [ ] My code follows the style guidelines (e.g., RuboCop)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally (`bundle exec rake test`)

### DIFF
--- a/lib/ancestry/class_methods.rb
+++ b/lib/ancestry/class_methods.rb
@@ -7,7 +7,7 @@ module Ancestry
       if object.is_a?(ancestry_base_class)
         object
       else
-        unscoped_where { |scope| scope.find(object.try(primary_key) || object) }
+        unscoped_where { |scope| scope.find_by!(primary_ancestry_key => object.try(primary_ancestry_key) || object) }
       end
     end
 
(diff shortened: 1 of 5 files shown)


kbrock:
@a5-stable does this look like it would meet your needs?

I'm trying to see if there is a way to patch this in or changes I can make in core so it is easier to patch.

Have gotten this to be less invasive, but it sure goes everywhere. I was hoping that I could localize the changes just to the builder. Haven't gotten there yet and a little stuck.

a5-stable:
@kbrock Hi, thanks for creating the PR! I tried running our product with this branch.
Ancestry works well and It seems to satisfy our requirements.
Is there anything I can do to help move this forward?

kbrock:
### Overview

I feel I am missing something.

I don't understand how this code can work without a primary id.

The existing STI support requires `unscoped` access, which will likely cause your plugin to fail.

My concern is I don't want to introduce a lot more complexity than I will support, and others will have to debug if I'm not sure it will benefit you.

Please help me understand how you associate 2 classes together and how the controller tells the model (and database) which record to return when you request `:user_id=>1`.

### Request

Do you have a simple example of a code base?

- 2-3(?) models.
- One has a hierarchy (that requires this branch).
- CRUD controllers/scaffolding with a show page.
- A view that displays the model/relationship at different time intervals.

### What I hope to gain

I want to look at the plugin/gem's use of default scopes or how the date is determined.

I want to see how you pass the data to the database so I can ensure the way Ancestry turns off scopes does not affect you.

Also, it seems like all records are read-only. Not sure how ancestry would handle this.

Thank you for your assistance.

PR #762, opened 2026-04-28
Docs: restore generated-column links and add version-tag note

sdhull:
## Summary

- The split of `CONFIGURATION.md` out of `README.md` in d8e65f2 dropped the link to MySQL's generated columns docs from the **Cached Columns** section and left the preceding code fence unclosed (the migration block at the end of "Migrating Between Formats" was never closed, so the `## Cached Columns` heading was being absorbed into a code block on GitHub's renderer). This PR restores the MySQL link, adds the equivalent [PostgreSQL generated columns](https://www.postgresql.org/docs/current/ddl-generated-columns.html) link, and closes the fence.
- Adds a note at the top of `README.md` and `CONFIGURATION.md` reminding readers that GitHub's default view is the `master` branch, which tracks active development. They should switch to the [tag](https://github.com/stefankroes/ancestry/tags) matching their installed gem version when consulting these docs — otherwise the docs may describe options or behavior that haven't shipped yet.

## Test plan

- [x] Verified the broken code fence is fixed (the `## Cached Columns` heading now renders as a heading, not code).
- [x] Confirmed both new links resolve (MySQL 8.0 generated columns, PostgreSQL current generated columns).
- [x] Re-rendered both files locally to check the blockquote note renders correctly above the existing content.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -5,6 +5,11 @@ Full reference for `has_ancestry` options, global defaults, and advanced setup.
 For getting started, see the [README](README.md).
 For upgrading between versions, see [Upgrading](docs/UPGRADING.md).
 
+> **Reading these docs on GitHub?** The `master` branch reflects bleeding-edge
+> development and may describe features not yet released. Use the branch
+> selector above to switch to the [tag](https://github.com/stefankroes/ancestry/tags)
+> matching the version of the gem you have installed.
+
 ## Table of Contents
 
 - [Ancestry Formats](#ancestry-formats)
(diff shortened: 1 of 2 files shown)


kbrock:
@sdhull thank you so much. If you find the changes confusing, please let me know and I can move stuff around to try and make it more clear.

The readme was just getting too long.

sdhull:
@kbrock no worries, I think splitting it up makes sense! I just noticed that one link to "Cached Columns" section wasn't working so figured I'd fix it! 

... and I admit I read about `ltree`, got excited about it and added a migration for it only to realize it's not supported in any released version yet 🤦 so I wanted to add that tip about making sure you're viewing the applicable version of the docs 😆

PR #770, opened 2026-09-20
Stop relying on the private validation_context= writer

### LINKED ISSUE #766: NoMethodError: undefined method 'validation_context=' on Rails main (removed in rails/rails#58220)
## Summary

The gem does not work with the Rails main branch after 2026-07-23. A `NoMethodError` occurs when you move a node that has descendants. The same error occurs when you run `check_ancestry_integrity!`.

## Environment

- ancestry: master (e8c09d2)
- rails: main (af95ef17)
- ruby: 4.0.1

## Cause

Rails PR rails/rails#58220 removed the private writer `validation_context=` from `ActiveModel::Validations`. The writer was not necessary in Rails. The method `valid?` sets the context on an internal `ValidationContext` object since rails/rails@afecc8fd6c. Only the reader `validation_context` stays public.

The method `sane_ancestor_ids?` assigns `self.validation_context`:

https://github.com/stefankroes/ancestry/blob/master/lib/ancestry/instance_methods_builder.rb#L103

Ruby permits this call for a private writer. The method is now fully removed. The call goes to `method_missing` and causes `NoMethodError`.

These code paths call `sane_ancestor_ids?` and fail:

- `_update_descendants_with_new_ancestry` (a move of a node that has descendants)
- `_check_ancestry_integrity!`
- `update_descendants_with_new_ancestry_sql`

## Steps to reproduce

1. Use Rails main after 2026-07-23.
2. Create this model:

```ruby
class Node < ApplicationRecord
  has_ancestry ancestry_format: :materialized_path2
end
```

3. Run this code:

```ruby
parent = Node.create!
child = Node.create!(parent: parent)
grandchild = Node.create!(parent: child)

child.update!(parent: nil) # move a node that has descendants
```

## Result

```
NoMethodError: undefined method 'validation_context=' for an instance of Node
  activemodel/lib/active_model/attribute_methods.rb:511:in 'method_missing'
  ancestry/lib/ancestry/instance_methods_builder.rb:103:in 'sane_ancestor_ids?'
```

`Node.check_ancestry_integrity!` causes the same error through `class_methods.rb:151`.

## Possible corrections

Option 1: Do not assign `validation_context` in `sane_ancestor_ids?`. The method runs the validators directly. A context change is possibly not necessary.

Option 2: Write to the new internal object when the old writer is not available:

```ruby
if respond_to?(:validation_context=, true)
  self.validation_context = context
else
  send(:context_for_validation).context = context
end
```

## Workaround

We use this shim until the gem changes:

```ruby
module AncestryValidationContextShim
  def validation_context=(context)
    context_for_validation.context = context
  end
end

ActiveSupport.on_load(:active_record) do
  include AncestryValidationContextShim
end
```

Thank you for the gem.

hikmetba-bit:
## Summary
Fixes #766.

`sane_ancestor_ids?` saved, nulled, and restored `self.validation_context` around a direct call to each column validator's `#validate_each`. Rails main removed the private `validation_context=` writer entirely ([rails/rails#58220](https://github.com/rails/rails/pull/58220)), since `ActiveModel::Validations#valid?` now sets the context on an internal `ValidationContext` object instead. The removed writer breaks `sane_ancestor_ids?` on current Rails main with:
```
NoMethodError: undefined method 'validation_context=' for an instance of Node
```
which in turn breaks moving a node that has descendants and `check_ancestry_integrity!` (both call this method), exactly as described in the issue (credit to the reporter for the excellent diagnosis, down to the exact Rails PR and line number).

## Fix
This is Option 1 from the issue: the save/restore was unnecessary to begin with. `validate_each` is called directly on each validator here, bypassing the normal `valid?`/`run_validations!` pipeline that actually *reads* `validation_context` — so nulling and restoring it around this direct call never affected which validators ran or how. Dropped both the write and the `ensure` restore.

## Test
Added `test_sane_ancestor_ids_does_not_rely_on_validation_context_writer` to `integrity_checking_and_restoration_test.rb`: it stubs `validation_context=` on a single instance to raise `NoMethodError` (simulating the writer's removal on Rails main) and confirms `sane_ancestor_ids?` still runs to completion. Mutation-tested: reverting the source change makes it fail with that exact `NoMethodError` through `sane_ancestor_ids?`; with the fix, the full `integrity_checking_and_restoration_test.rb` file passes (4/4). Ran the full suite (`SKIP_NATIVE_DB_GEMS=1 bundle exec rake test` — trilogy/pg require native client libs not available in this sandbox, so I tested against the default SQLite adapter only): 259 runs, same 5 pre-existing `SortByAncestryTest` failures with or without this change (confirmed by reverting and re-running - those failures are SQLite ordering-related and unrelated to this fix), no new failures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kbrock:
@hikmetba-bit so we are using this to detect if we introduced an error or if there was an error from the model. (some people save records to the database with errors in them - which happens when you modify your validations and don't revalidate every record in the database)

So this concept is necessary in general Maybe there is a better way for us to validate that we properly updated the ancestry with a custom validator or calling the validator directly?
I was trying to get away from custom logic and go towards leveraging the validators that are already setup.

Need to look into how rails suggests running a subset of validators. I think this concept is still necessary to rails in general, so it may be as simple as tracking down how this changed.

